PR #10, opened 2019-08-12
Data

Farnazmdi:
@aarmey I added a function to remove the outlier (peaks) to the `importData.jl`. It works well for the now, I have checked for all trials of three drugs, but maybe later I should replace it with a more general outlier removal function. I couldn't find any function in Julia itself yet for it. Temporarily I have added a new notebook to show the output of the function; once you confirm the function is ok, I will remove the [notebook](https://github.com/meyer-lab/one_state_model/blob/data/model/remove_peaks.ipynb).

here the `remove_peaks()` is the function
[here](https://github.com/meyer-lab/one_state_model/blob/e3475f8287f9a2c135fc785183231f0e552c1cc0/model/importData.jl#L40)


Farnazmdi:
To resolve the conflicts, should I just follow the lines that here is suggested? 
image

aarmey:
@Farnazmdi click on resolve conflicts. Each conflict includes two blocks of code—the parts from master and data within the conflicting part. You want to make the file like what you want to keep.

Farnazmdi:
@aarmey  Thanks, seems to be resolved.

PR #27, opened 2019-08-21
Enhance

Farnazmdi:
This one in the enhance branch is up to the last step that all the b parameters for Hill are the same, and it was working good.
I think we could merge this for now if you agree. I have been playing with the new changes in the branch improve, to make taus proportional to rate parameters, but the fit doesn't seem to be good. I have tried many algorithms for solving, but it is not as good as the one in the last commit even for one concentration. But I'll keep trying.


data/Parameters_D.csv
```diff
@@ -0,0 +1,7 @@
+x1,x2,x3,x4,x5,x6,x7,x8
+0.05891223934526264,0.07729363212423296,0.13250621348370295,0.010813664989686791,0.024483796962931292,0.002480384884092555,0.0041040221876294915,0.03584929065117992
```
aarmey: I'd rather not store these in the repo, because they should be derived from the original data.

src/python_files/cell_utils.py
```diff
@@ -58,17 +59,22 @@ def generateLineageWithTime(initCells, experimentTime, locBern, g1_a=None, g1_b=
...
     # have cell divide/die according to distribution
     for cell in lineage:
+<<<<<<< HEAD
+        cell.g1 = sp.gamma.rvs(g1_a, scale=g1_b)
+        cell.g2 = sp.gamma.rvs(g2_a, scale=g2_b)
+=======
         cell.g1 = sp.gamma.rvs(g1_a, scale = g1_b)
         cell.g2 = sp.gamma.rvs(g2_a, scale = g2_b)
-        
+>>>>>>> ecc2d2f190bd2c4614836e2258346e7e847da2db
```
aarmey: Fix merge conflict errors.

Farnazmdi:
@aarmey  one quick question, should the concentrations in the hill function be necessarily log scaled?

aarmey:
> @aarmey one quick question, should the concentrations in the hill function be necessarily log scaled?

Because you have them up in the exponent of your hill equation, yes. If you divide the concentration by the EC50, then no: https://en.wikipedia.org/wiki/Hill_equation_(biochemistry)

aarmey:
P.S. I've given you admin access to the repo, so you should be able to override and merge even if codecov complains.

Farnazmdi:
Oh, so I have to convert the concentrations...
Thank you, it seems if I write the tests it would complain less :))

Farnazmdi:
Just make the upper and lower bound a little tighter, and plotted all the concentrations for lapatinib in the `hill.ipynb`. The fit is a little better now.

PR #33, opened 2019-09-16
Farnaz

Farnazmdi:
This PR includes the changes to the Hill fitting (not the DDE) to set the delays as constant. 
The tests have been changed accordingly and they passed. Now the Hill has 10 parameters (instead of 12) to estimate.


Farnazmdi:
There are many changes since I ran the notebook again and deleted some notebooks (which were addressing the same simulation for other drugs separately and I decided to get the final result with just lapatinib and then apply the same thing for other drugs). But actually I just changed the [hill function](https://github.com/meyer-lab/DrugResponseModel.jl/blob/8bcd5cb45f3903695e10c7cf6f60af8feab5ba96/src/Hill.jl#L23) and in its corresponding [notebook](https://github.com/meyer-lab/DrugResponseModel.jl/blob/farnaz/src/hill-lapatinib.ipynb)

aarmey:
This all looks good. Do the notebooks run for you, though? I'm getting some errors but am note sure from where. If they run for you then merge as it's probably just something on my user account.

Farnazmdi:
Thank you, you are right the `plots-jl.ipynb` hadn't been changed accordingly, Now all the notebooks should work, and I deleted those notebooks that are not being used for now.

PR #34, opened 2019-09-17
plots for the poster 

Farnazmdi:
In this PR I just changed the `plots_jl.ipynb` and a few other changes to the files to make the poster plots.


Farnazmdi:
Dr. Meyer, I switched to `pyploy()` instead of `gr()` for the plots, there is an error regarding the installation of `matplotlib` 
image
I tried the suggestion here, but the command was not found. Also, I have matplotlib in my python3  I couldn't figure out what to do. @aarmey

aarmey:
I'm not sure how to handle this. What is the problem with gr?

Farnazmdi:
I had a problem with the legend font, but that doesn't seem to be a problem anymore, I will change it back to gr().

PR #38, opened 2019-09-20
function for the phase portrait

Farnazmdi:
I added a function to plot the phase portraits and updated the notebook, 
I added it to the last plot here in the [notebook ](https://github.com/meyer-lab/DrugResponseModel.jl/blob/correlation/src/plots_jl.ipynb).


Farnazmdi:
I'll add some more tests to cover the plot functions in another branch, I guess the code coverage is about them.

PR #42, opened 2019-09-24
Hill improved significantly

Farnazmdi:
Made few changes in the direction of using the same function everywhere, but more importantly, I caught my mistake in the Hill, that I had to use the log transform of the parameters and then pass it to the `ddesolve`. The fitness function has significantly reduced (from 600,000 to 12,000). Will try to more improve it in another branch.


Farnazmdi:
image

Farnazmdi:
image
This is the best I could get.

Farnazmdi:
image

Farnazmdi:
I returned the tau1 and tau2 to hill function, doesn't matter very much.

Farnazmdi:
If this Hill fitting is acceptable, I think this [notebook ](https://github.com/meyer-lab/DrugResponseModel.jl/blob/dde/src/plots_jl.ipynb)is ready if we want to show it to Dr. Heiser and Sean. @aarmey

aarmey:
This looks great! We can keep all the parameters as a Hill curve—it seems to be working well here.

src/DDEmodel.jl
```diff
@@ -80,9 +76,4 @@ function optimization(g1, g2, g1_0, g2_0, initial_guess, j, lower, upper, num_st
     println(best_fitness(results_dde))
     new_guess = best_candidate(results_dde)
     return best_fitness(results_dde), exp.(new_guess)
-#     println("local optimization begins")
-#     optim_res = optimize(obj, lower, upper, new_guess, Fminbox(), Optim.Options(show_trace=true))
-#     println("the fitness after local optimization : ")
```
aarmey: Do you plan to add these lines back? Delete if not.

aarmey:
(Let me know if you want me to override codecov.)

src/DDEmodel.jl
Farnazmdi: It (surprisingly) didn't make an improvement, I will delete it.

Farnazmdi:
Thanks, I had forgotten to uncomment some tests, it should be fine now. But if it could help with not giving an error because of the long-time run, that would help a lot.

aarmey:
Actually, it would be best to ensure the tests can run to completion. Maybe now that things seem to work, you only need to run fitting once with one drug? The idea of tests is just to check that nothing is **obviously** broken.

Farnazmdi:
In the current files, I have tests for all drugs and I check their fitness to make sure they are not bad, and they all have 5 iterations. This was based on one of the issues. Do you mean I just run them for 1 iteration? Or just test for one drug?

aarmey:
Limiting to one iteration might be enough to speed things up.

aarmey:
Sorry, I was thinking iteration meant you tried the full fitting process multiple times. You still want iterations during fitting. I guess limit to one drug.

Farnazmdi:
No worries at all, I'll do it right away.

PR #57, opened 2019-11-08
drug combination

Farnazmdi:
@aarmey  I added a combination function [here](https://github.com/meyer-lab/DrugResponseModel.jl/blob/3cdbfb2a82e673b8998cb6078063f2b1959f1adb/src/Hill.jl#L70):
Is this intended as a drug combination?


Farnazmdi:
The dosage amounts for different drugs are not on the same scale (lapatinib goes from 0-500 nM, gemcitabine from 0-100 nM) I am not sure if this is an issue, but for all drug concentrations, I am summing their halves to get the combination effect.

aarmey:
Nope. Look up the [Bliss independence](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6415926/) model. It should be calculated that sort of way.

Farnazmdi:
@aarmey  quick question. Here we have 6 drug effects. Should we just consider gamma_1 and gamma_2 (like the inhibition effect mentioned in that paper) for the combination effect? or all the 6 rates?

aarmey:
Think you have to handle all 6 rates.

Farnazmdi:
Ok thanks, one more question. I think I need to normalize them to be between zero and 1, right? For delays, it becomes negative if I keep them in their own units when calculating Bliss independence.

aarmey:
Yep, Bliss only works scaled 0-1. So, you might want to only try it with the alpha and beta parameters first. I think it might be reasonable to assume the delays are constant. For the death rates, you could assume the effect is additive (death_combination = death_1 + death_2).

Farnazmdi:
image

Farnazmdi:
And this is for gammas after adding them together as total death rate
image

Farnazmdi:
The functions added are 
[these](https://github.com/meyer-lab/DrugResponseModel.jl/blob/03e9db272bc5b4be2a7e534d915505f15cb42462/src/Hill.jl#L70)  two functions. I kept the 3d plotting function inside the notebook because there were some settings for PyPlot() which were giving an error. I hope this is the way it should be!

src/Hill.jl
```diff
@@ -66,3 +66,31 @@ function getDDEparams(p, concentrations)
...
+    """ A function to calculate Bliss independence for drug combination assuming
+    the two drugs hit different pathways and they effect completely independently. """
+
+    param1 = ParamForBliss(p1)
+    param2 = ParamForBliss(p2)
+    Effect = zeros(8,8,3)
+    for i in 1:3
+        for j in 1:8
+            for k in 1:8
+                Effect[j,k,i] = param1[i,j] + param2[i,k] - param1[i,j] * param2[i,k]
```
aarmey: Indentation doesn't seem to line up for these loops.

Could get rid of the i loop and vectorize this:

`Effect[j,k,:] .= param1[:,j] .+ param2[:,k] .- param1[:,j] .* param2[:,k]`

PR #58, opened 2019-11-18
Plot combination

Farnazmdi:
plotting the DDE time series for the combination of lapatinib and gemcitabine


Farnazmdi:
@aarmey  In this [notebook](https://github.com/meyer-lab/DrugResponseModel.jl/blob/plot/src/hill.ipynb) I brought the 64 figures from the combination of the Lapatinib and gemcitabine.
Please let me know if you think they should be in another way, or you want them different.
I would like to include them in my slides for M&M meeting on Wednesday.

aarmey:
These look reasonable. You'll want to include the concentration of both drugs each plot corresponds to. Also, be sure to walk us through the interesting aspects rather than all the plots.

Farnazmdi:
I'm keeping the plot functions inside the notebook because `pyplot()` seems to have some problem being installed, but the figures with pyplot are much better to handle. 

One quick sanity check; since I just had 3 parameters as Bliss combination (alpha, beta, gamma as the sum of gamma_1 and gamma_2), I used gamma_1 and gamma_2 = gamma/2, and the average of other parameters for other ones, like [here](https://github.com/meyer-lab/DrugResponseModel.jl/blob/0d76b5e57eb3b2ade33f74be16143a81276bdfee/src/Hill.jl#L101) and the following a few lines. Does this make sense? @aarmey

Farnazmdi:
And I can't access the `conflicts` to resolve, I can't find them in the file. I tried the manual one but it says:
image

aarmey:
Just do `git checkout plot` instead of that line.

aarmey:
Don't understand your description of the parameters. alpha and beta should be through Bliss independence. gamma terms should be the added term for each individual drug. You can just keep the delays constant.

Farnazmdi:
Got it, thanks.

PR #66, opened 2019-12-13
ODE 4eq

Farnazmdi:
I will figure out how to just have one function for the 2-, 4- (and maybe 6) equation model. For now, they are in a different file.


src/ODE4.jl
```diff
@@ -0,0 +1,83 @@
...
+    G1 = solution(t, idxs=1).u + solution(t, idxs=2).u
+    G2 = solution(t, idxs=3).u + solution(t, idxs=4).u
+    res[1, :] = (G1 - g1[:, i]).^2
+    res[2, :] = (G2 - g2[:, i]).^2
+    summ = sum(res[1,:]) + sum(res[2,:])
+    return summ
+end
+
+""" Fit the ODE model to data. """
+function ODEoptimizer4(lower_bound::Array, upper_bound::Array, p::Array, i::Int, g1::Matrix, g2::Matrix, g1_0::Array, g2_0::Array)
```
aarmey: You shouldn't need to change the lower and upper bounds for the problem, so I'd set them in the function here, rather than pass them as arguments.

src/ODE4.jl
```diff
@@ -0,0 +1,84 @@
...
+end
+
+function ODEmodel4(p)
+    @assert all(p .>= 0.0)
+    return DiffEqArrayOperator([-p[1] .- p[5] 0.0 0.0 2.0*p[4]; p[1] -p[2] .- p[5] 0.0 0.0; 0.0 p[2] -p[3] .- p[6] 0.0; 0.0 0.0 p[3] -p[4] .- p[6]], update_func=update_coef4)
+end
+
+""" Predicts the model given a set of parametrs. """
+function predict(p, g1_0, g2_0, i, t)
+    u0 = 0.5*[g1_0[i], g1_0[i], g2_0[i], g2_0[i]]
```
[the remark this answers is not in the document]
Farnazmdi: Thanks, addressed.

src/ODE4.jl
Farnazmdi: done.

PR #141, opened 2020-01-15
test and bound range

Farnazmdi:


Farnazmdi:
@aarmey I understand that if nD1 or nD2 is zero, then we cannot have `ones(nD1)` or the other one and that' why it errors now. 
Should I use and if statement to say if nD1 and nD2 == 0 the jacobian is in another form?

aarmey:
Yes, you'll probably just have to handle this edge case with an if statement.

Farnazmdi:
@aarmey It hasn't got any better. Although now zero is allowed for death steps, it doesn't find it as optimum.

aarmey:
Then there must be an error somewhere.

I'd take a previous parameter set that worked for the model without death delay, and add a test that it still provides a low error with `nD1=0` and `nD2=0`. With that you can be confident your code is right. From there, then, you can play with the fitting process.

src/ODEmodel.jl
```diff
@@ -31,25 +48,50 @@ function predict(p, g1_0::Real, g2_0::Real, t, nG1::Integer, nG2::Integer, nD1,
...
 
     G1 = Vector{eltype(p)}(undef, length(t))
     G2 = Vector{eltype(p)}(undef, length(t))
 
-    for ii = 1:length(G1)
-        G1[ii] = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
-        G2[ii] = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
-
-        v = A * v
+    if nD1 == 0 & nD2 == 0
```
aarmey: `A = zeros(0)` doesn't work

But `A = []` gives you an empty vector.

Farnazmdi:
Thanks.

Farnazmdi:
image
@aarmey I don't see where this error is occurring. Could you please take a look at it? It says when nD1 or nD2 = 0, this happens, but I couldn't find any `zero` as an incorrect syntax anywhere...

aarmey:
@Farnazmdi see last commit. Apparently empty arrays still need to have a type, or functions get confused. Note that you should read a stack trace from bottom up. You don't call `zero()`, but `diagm()` calls `zeros()` which calls `zero()`.

src/ODEmodel.jl
```diff
@@ -5,21 +5,43 @@
...
     A[nG1 + nG2 + 1, 1:nG1] = p[3] * ones(1, nG1)
     A[nG1 + nG2 + nD1 + 1, (nG1 + 1):(nG1 + nG2)] = p[4] * ones(1, nG2)
 
     rmul!(A, dt)
 
-    @assert all(A[1:(nG1 + nG2), (nG1 + nG2 + 1):end] .== 0.0)
-    @assert A[nG1 + nG2 + nD1 + 1, nG1 + nG2 + nD1] == 0.0
-    @assert all(A[(nG1 + nG2 + 1):(nG1 + nG2 + nD1), (nG1 + nG2 + nD1 + 1):end] .== 0.0)
-
+    if nD1 & nD2 != 0
```
aarmey: Why do you need this condition?

src/ODEmodel.jl
```diff
@@ -5,21 +5,43 @@
...
+        diagD2 = -ones(nD2) * p[4]
+    end
+
+    v1 = [-ones(nG1) * (p[3] + p[1]); -ones(nG2) * (p[4] + p[2]); diagD1; diagD2]
+    v2 = [ones(nG1) * p[1]; ones(nG2 - 1) * p[2]; D1; D2]
+    A = diagm(0 => v1, -1 => v2)
 
     A[1, nG1 + nG2] = 2 * p[2]
     A[nG1 + nG2 + 1, 1:nG1] = p[3] * ones(1, nG1)
     A[nG1 + nG2 + nD1 + 1, (nG1 + 1):(nG1 + nG2)] = p[4] * ones(1, nG2)
```
aarmey: Shouldn't nD2 appear in this?

src/ODEmodel.jl
Farnazmdi: because if they are zero, then `+1` in the indexes goes beyond the matrix dimensions I guess.

src/ODEmodel.jl
Farnazmdi: I have both of them appear in the `if statement` above this, so I just kept D1, D2, diagD1 and diagD2

Farnazmdi:
@aarmey has anything updated in the server? I can't run any of the notebooks in my local.

Farnazmdi:
Please ignore my question. It is working now, but for almost 2 hours I was trying and it just didn't run.

aarmey:
Huh. Don’t know if any relevant changes. Let me know if it happens again and I can check a few things. Was it hanging, was the kernel crashing, or something else?

Farnazmdi:
Thank you,
It looked like when we do "restart the kernel and clear all outputs", and when I wanted to restart to run the cells, it just didn't do anything and the kernel just said it is restarting for a loooong time.

aarmey:
Found the problem... looks like Marc is fitting 1000 models at once 😂. I'll see if he can run them a bit more sequentially.

aarmey:
You also have a build here that's been running forever.

aarmey:
Time for another server...

Farnazmdi:
I stopped the last one and built again, but you are right this one took a long time, too.
Do you mean I just rebuild it, or is it possible to manually change the server?

aarmey:
I'm going to be getting another machine that's about 2X as fast. Then we can also split up users.

Farnazmdi:
The only thing that doesn't look better is paclitaxel trial 6. Although the model has the option to choose zero for the death number of steps, it is still choosing more than 1. I guess it is good enough. Please let me know if you think it needs more modifications.

Farnazmdi:
It is running 2 of the notebooks, that's why it is taking 1 and a half hours. I can just set the parameters so it doesn't need to run the optimization to build. Should I do that or just leave it this way?

aarmey:
Yep, if you have a good parameter set please set them. We can look into speeding up the optimization later, but I'd rather get some results with combinations, etc, first.

Farnazmdi:
Just need your approval for merge

aarmey:
Please remove output from lapatinib notebook.

PR #151, opened 2020-02-03
Combination

Farnazmdi:
2 functions added.
1. `combination`, that finds the number of cells in G1 and G2, given a set of ODE parameters for a specific drug concentration. It creates a paramRange (like the one we had in sensitivity). Returns the # cells in g1, # cells in g2, and the parameter range.
2. `central_difference`. I wrote the central difference function to get the derivative of the number of cells in each phase with respect to the parameter range.

I guess we need to define a threshold. I plotted them almost all, for alpha, beta, gamma1, and gamma2. The behavior mostly makes sense except for beta that has 1 or 2 small jumps at the beginning.

### DIFF
--- a/src/Hill.jl
+++ b/src/Hill.jl
@@ -135,32 +135,31 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
     ylims!((1E2, 1E4))
 end
 
-""" Calculate the gradient of # of cells in each phase w.r.t. a parameter. """
-function combination(params, i, g0)
+""" Calculate the # of cells in G1 for a set of parameters and T """
+function numG1(params, g0, T)
     t = LinRange(0.0, 95.5, 192)
-    convRange = 10 .^ (range(-1, stop = 1, length = 101))
-    paramRange = params[i] .* convRange # params is ODE parameters
-    g1 = zeros(length(paramRange))
-    g2 = zeros(length(paramRange))
+    temp = copy(params)
+    G1, _ = predict(temp, g0, t, Int(floor(temp[6])), Int(floor(temp[7])), Int(floor(temp[8])), Int(floor(temp[9])))
 
-    for j = 1:length(paramRange)
-        temp = copy(params)
-        temp[i] = paramRange[j]
-        G1, G2 = predict(temp, g0, t, Int(temp[6]), Int(temp[7]), Int(temp[8]), Int(temp[9]))
-        g1[j] = G1[120]
-        g2[j] = G2[120]
-    end
-    return g1, g2, paramRange
+    return G1[T]
 end
 
-""" Calculate central difference. """
-function central_difference(g1e, g2e, paramRange)
-    @assert length(g1e) == length(g2e) == length(paramRange)
-    der_g1 = zeros(length(paramRange) - 1)
-    der_g2 = zeros(length(paramRange) - 1)
-    for i = 1:length(paramRange) - 1
-        der_g1[i] = (g1e[i+1] - g1e[i])/(paramRange[i+1] - paramRange[i])
-        der_g2[i] = (g2e[i+1] - g2e[i])/(paramRange[i+1] - paramRange[i])
-    end
-    return der_g1, der_g2
+""" Calculate the # of cells in G2 for a set of parameters and T """
+function numG2(params, g0, T)
+    t = LinRange(0.0, 95.5, 192)
+    temp = copy(params)
+    _, G2 = predict(temp, g0, t, Int(floor(temp[6])), Int(floor(temp[7])), Int(floor(temp[8])), Int(floor(temp[9])))
+
+    return G2[T]
+end
+
+""" Calculates the gradient with central difference"""
+function diffG1(params, g0, T)
+    difG1(x) = numG1(x, g0, T)
+    return Calculus.finite_difference(difG1, params)
+end
+
+function diffG2(params, g0, T)
+    difG2(x) = numG2(x, g0, T)
+    return Calculus.finite_difference(difG2, params)
 end
--- a/src/Hill.jl
+++ b/src/Hill.jl
@@ -161,8 +161,10 @@ function plotGradient(effects, concentration, g0, T)
         dif[:, i] = diffCell(params, g0, T)[1:4]/num
     end
     concentrations = log.(concentration)
-    plot(concentrations, dif[1,:], label="alpha", lw=2, legend=:bottomright, xlabel="log concentration [nM]", ylabel="gradient of #cells wrt param")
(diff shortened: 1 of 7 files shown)


aarmey:
Can you post what you plotted? I'm not following what you did here.

You definitely don't want to implement central differencing yourself. Use `Calculus.finite_difference()` or `FiniteDiff.finite_difference_gradient()`

Farnazmdi:
This is the # of cells in G1 vs. a range of gamma2.
image

And this is the derivative of above w.r.t. gamma2:
image

The problem with using `Calculus` was it needed a function, but I just had data points. I tried to make a function but it didn't work, so I just wrote what it was on the Wikipedia. I notice it needs a little bit of change though fo the x-axis.
Is this wrong?

aarmey:
But you just need a function that takes in a set of parameters, and outputs the number of cells at `t=T`. Isn't `predict()` 99% of that?

aarmey:
I think you're differencing the same parameter with different intervals... not sure how to use that. For example, what is the x-axis on your plot? What does that mean biologically?

Farnazmdi:
Yes sorry, you are right. This is useless. I thought the x-axis indirectly will tell us about the concentration. 

What we want is to get the partial derivative of a function -- that is a function of all parameters and `T` -- with respect to one of the parameters.
Is it correct?

aarmey:
Yep.

src/Hill.jl
```diff
@@ -134,3 +134,32 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
     plot!([realParam], seriestype = "vline", margin = 0.3cm, legend = :false)
     ylims!((1E2, 1E4))
 end
+
+""" Calculate the # of cells in G1 for a set of parameters and T """
+function numG1(params, g0, T)
```
aarmey: Except you don't want numG1 or numG2, you want number of total cells. You should just have one function with these added together.

Farnazmdi:
Should I calculate this for all the combinations of the parameters?

aarmey:
We should look at a plot with drug concentration on the x-axis, and d[#]/d[alpha], d[#]/d[beta],  d[#]/d[gamma1], and d[#]/d[gamma2] as the y-axis quantities.

Farnazmdi:
doxorubicin:
image

lapatinib:
image

gemcitabine:
image

paclitaxel:
image

aarmey:
First, concentration should be a log axis. Second, looks like alpha/beta should probably be on a separate plot. Third, it might work better to plot d[number]/d[param]/number. I.e. normalize the derivative to the number of cells, so you’re looking at the proportional effect.

Farnazmdi:
Sorry I bothered you a lot in this PR.
I couldn't figure out how to interpret this. 
I understand that we want the combination of the two drugs that one of them decreases gamma1 (for example) and the other decreases gamma2. By decrease, I mean more negative in this plot. The more negative the gradient is, the more the cells are dying? 
I guess `beta` could be seen as the division rate, but do we want the same thing as above with the transition rate, too? I mean do we want to pick those two drugs that one decreases `alpha` and the other decreases `beta`?

Sorry I can't get a sense of the derivative of a function -- with some parameters -- that returns the number of something, with respect to one of its parameters.
I have this analogy: 
The velocity is a function of time, and when we write it as a linear function, acceleration is one of the parameters in the formula.
The number of cells is a function of concentration, but we have transition and death rates as its parameters. 
I don't get what is meant by derivative of velocity wrt the acceleration.


doxorubicin:
image

gemcitabine:
image

lapatinib:
image

paclitaxel:
image

aarmey:
I don't follow your velocity vs acceleration analogy. Velocity is the first derivative, acceleration is the second derivative. We have no second derivatives here.

These are the direction of change effect on cell # when increasing the indicated parameter.

Farnazmdi:
Yes sorry, it is not a good example. I meant v(t) = at + v0 ==> dv(t)/da.

So it means for each concentration (or a set of parameters), when we increase a parameter by one unit, how this affects the cell #. Is that what it is?

aarmey:
I think that's a reasonable description. So, for example, it looks like you should combine lapatinib with a drug that induces gamma1, not gamma2. Or, a drug that inhibits alpha over one that inhibits beta.

Farnazmdi:
Thank you.

src/Hill.jl
```diff
@@ -134,3 +134,38 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
     plot!([realParam], seriestype = "vline", margin = 0.3cm, legend = :false)
     ylims!((1E2, 1E4))
 end
+
+
+""" Calculate the # of cells in G1 for a set of parameters and T """
+function numcells(params, g0, T)
+    t = LinRange(0.0, 95.5, 3)
+    temp = copy(params)
```
aarmey: Is there a reason you need to copy params?

src/Hill.jl
```diff
@@ -134,3 +134,38 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+    diffcells(x) = numcells(x, g0, T)
+    return Calculus.finite_difference(diffcells, params)
+end
+
+
+""" Plot the gradient vs concentrations """
+function plotGradient(effects, concentration, g0, T)
+    dif = zeros(4,8)
+    for i = 1:8
+        params = effects[:, i]
```
aarmey: Why copy out effects into the `param` variable?

src/Hill.jl
```diff
@@ -134,3 +134,38 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+        num = numcells(params, g0, T)
```
aarmey: Why not do `numcells` correction inside `diffCell()`?

src/Hill.jl
Farnazmdi: Oh this has remained from before that I wanted to sweep the parameters. I will remove it.

src/Hill.jl
Farnazmdi: I will correct it.

src/Hill.jl
Farnazmdi: Will do right away.

Farnazmdi: Wait, sorry what correction? I guess I need the function. You mean I use `predict` as the function to pass to `Calculus...`?

aarmey: I can do this after you do the others. Just ping me.

Farnazmdi: Sorry, ignore it.

Farnazmdi: I made the edits. I took the normalization into the `diffCell()` but I'm not sure this addresses your point.

Issue #152, opened 2020-02-05
Predict for one timepoint


aarmey:
Write a version of `predict()` that only takes one timepoint, then also make `numcells()` use this version of `predict()`. There are much faster versions of the matrix exponential when you only need the answer once like this (Krylov approximation), and they work with autodiff. I can change the solving method once you have this structure in place. Rather than using central differencing, we can use ForwardDiff for your gradients after this change. This should be faster and more accurate.

PR #153, opened 2020-02-05
Predict2.0

Farnazmdi:


src/ODEmodel.jl
```diff
@@ -84,6 +84,29 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2)
...
+    else
+        D2 = zeros(nD2)
+    end
+
+    v = [ones(nG1) * p[5] * g_0 / nG1; ones(nG2) * (1.0 - p[5]) * g_0 / nG2; D1; D2]
+    A = ODEjac(p, t[2], nG1, nG2, nD1, nD2)
+    w = expmv(t[2], A, v)
+
+    return sum(w)
+end
```
Farnazmdi: I am using `expmv()` in Expokit. This is giving me a very large number as the total number of cells. I know there is something wrong. Could you please just let me know if I'm on the right track with `expmv()`? @aarmey

Farnazmdi:
#152

src/ODEmodel.jl
Farnazmdi: Oh I caught one. `ODEjac` returns the exponential matrix, and I'm exponentiating again.

aarmey:
Ok, I'll take this from here.

aarmey:
Ok, made the changes here. I didn't do the notebooks, so please fix those. You just pass the time now, not a vector with the time in the second element.

Farnazmdi:
Are we using any preallocation in `ODEjac`?
There is this error about that `ODEjac` cannot take Dual numbers:
image


I read that AD doesn't work with preallocated functions

aarmey:
Oops. Think I just fixed that.

Farnazmdi:
`diffCell` still doesn't work
image

I found this:
[link](https://github.com/JuliaLang/julia/issues/26552)
It looks like there is nothing to work around it

Farnazmdi:
Is this way more accurate than calculating all the points?
The behavior is less monotonic.

aarmey:
Post plot?

Farnazmdi:
dox:
image

gemcitabine:
image

lapatinib:
image

paclitaxel:
image

aarmey:
Hmm... the output should be the same. I did change the time since you were indicating the 100th timepoint before, and now it's 100.0 hrs. Maybe you could figure out what time the 100th point is, and then set it to that?

aarmey:
Also adding more concentrations may help find the issue. Because there's oscillatory behavior the result doesn't have to be monotonic, but it should be locally smooth.

Farnazmdi:
Of course. Thank you.

Farnazmdi:
tmp
The animation looks cool! This is the plot for almost all times.

Farnazmdi:
Could I keep it in the notebooks?

Farnazmdi:
Here is the final version:
dox:
dox

paclitaxel:
pac

lapatinib:
lap

gemcitabine
gem

aarmey:
The animation is neat, but all that matters is the result at 96+ hrs. Are you sure that's right?

And can you add more points along the x-axis? Like I mentioned before, we should make sure the result makes sense, and one way to check is that it is locally smooth.

aarmey:
E.g. how does increasing gamma1 with dox result in more cells?

Farnazmdi:
> E.g. how does increasing gamma1 with dox result in more cells?

If I'm not mistaken you are pointing to the peak in gamma 1, I guess it might have something to do with the normalization. Because cell # oscillates, I think it causes the peak and valley in the curve. If we look at the not-normalized plot, we don't see that:

tmp (1)

aarmey:
Bigger picture—do these results make sense? You need to convince yourself whether or not these do. For example, in this last graph, how would increasing gamma1 *ever* increase the number of cells?

Farnazmdi:
Although I think the parameters should be independent, but maybe because for lower concentrations alpha is large, a lot of cells are transitioning to G2. However, it shouldn't give rise to the number of cells, because they are transitioning, not dividing. It doesn't make sense.
Oh I guess I messed something up here:
image

I guess I am considering those that are in death phases as alive cells...
Could it be it?

Farnazmdi:
Still, beta is almost constant, shouldn't increase the cell number... right?

Farnazmdi:
One question, the reason that we are interested in 96+ hours is that more realistically, the treatment runs for more than 4 days? 

Also, if I'm not mistaken, for now with this approach, we just want to identify those drugs that are good together more than individually, and then we will work on the optimized concentration of each of them. Is this correct?

aarmey:
Right. Only the long-time behavior is meaningful for treatment.

And yes, ideally we'd like to figure out how drugs given together behave.

Farnazmdi:
@aarmey  Wouldn't it be alright to just use cell # versus concentration, for a range of each parameter, instead of the gradient? I mean, what is the advantage of the gradient over just cell number?

aarmey:
All drugs decrease cell # over concentration, the question is how. The gradient tells you what effect you want to use to decrease cell # even further.

Farnazmdi:
image
@aarmey I couldn't figure out what the problem is. Could you please let me know what I did wrong or what I should do to correct this?

aarmey:
This is related to my post on #general. Should be fixed now, but juptyerlab won't be available until I hear back from seasnet.

Farnazmdi:
Thanks.

Farnazmdi:
Sometimes the cell number becomes negative. I put an assertion for that, I will try to see if this used to happen before the change to the jacobian.

Farnazmdi:
I guess there is something wrong with the `Calculus.finite_difference()` because as it is clear here, with exactly the same set of parameters, the number of cells is decreasing.
image

The y-axis is the number of cells, and the x-axis is a range of gamma_1.
image

@aarmey do you see any other point that might be going wrong here?

Farnazmdi:
Something that looks weird is that with a higher concentration (this is in gemcitabine) the G2 rate progression -- which also happens to be division rate -- increases! I searched a little to see if I could find some evidence of this, but couldn't.

Farnazmdi:
Actually, there is something wrong with the `predict2()`, it sometimes returns negative cell numbers.

Farnazmdi:
Since the solution in `predict()` and `predict2()` is not the same, and the second one sometimes returns negative cell numbers, if you let me I will use the `predict()` and just take 1 time point. Is that alright?

aarmey:
Yep. In that case get rid of predict2

Farnazmdi:
Some updates:
1. paclitaxel and doxorubicin seem to work alright in terms of their gradient. Gemcitabine goes positive for two last concentrations. There should be something related to the fact that beta increases significantly over concentrations. 

There are two errors now.
1. For all of the sensitivity analyses, the only first cell # is some negative small number (like -0.002). I changed the range we have for the parameters in sensitivity analysis, but it just happens with any range.
2. In lapatinib, the number of death phase steps in G1 (nD1) is estimated to be zero. This messes with something that it errors for `invalid array dimensions`. It occurs when it reaches the line here:

image

Farnazmdi:
This is the error regarding `invalid array dimensions`.
image

Farnazmdi:
@aarmey  I can't figure out what array dimension is wrong. Could you pleeease take a look at this error? When nD1 is estimated to be 0 for lapatinib, this error happens. I mean when the matrix dimension for death in G1 becomes zero. The fitie_difference breaks in this. I tried a few tricks to just remove that parameter, but it breaks other things. I would really appreciate if you could take a look at this. Any recommendations helps. 
image

src/ODEmodel.jl
```diff
@@ -74,21 +74,39 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2)
     G1 = Vector{eltype(p)}(undef, length(t))
     G2 = Vector{eltype(p)}(undef, length(t))
 
-    for ii = 1:length(G1)
-        G1[ii] = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
-        G2[ii] = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
+    if nD1 == 0 & nD2 != 0
```
aarmey: Why do you need an if statement here? If nD1 is equal to 0 doesn't it just end up having no effect on addition?

aarmey:
If we're not ultimately using predict2, then what do we need to keep from this branch?

src/ODEmodel.jl
Farnazmdi: In either of G1 or G2, the second term `sum(...)` would be meaningless, for example in G1, since it is summing the index `nG1+nG2+1 : nG1+nG2` if nD1= 0.

Farnazmdi:
I was trying to address the issues in the gradient here, but you are right, I could do it in another branch.

Farnazmdi:
This error I mentioned still exists even with what it is in the branch master. It is for the control trial in lapatinib.

aarmey:
Yes, if there's nothing you need here, it'd be helpful for debugging to make a failing case off of master.

PR #162, opened 2020-02-23
tests

Farnazmdi:
I added a test in `testHill.jl` at the very last line to test the gradient function for a set of parameters in which the nD1 is zero, and it fails.

Apart from this, for now, just Gemcitabine doesn't behave expectedly in terms of the gradient. I guess the notebooks should run now.

### DIFF
--- a/src/Hill.jl
+++ b/src/Hill.jl
@@ -140,8 +140,8 @@ end
 function numcells(params, g0, T)
     t = LinRange(0.0, 200.0, 201)
     G1, G2 = predict(params, g0, t, Int(floor(params[6])), Int(floor(params[7])), Int(floor(params[8])), Int(floor(params[9])))
-    @assert(all(x -> x>=-0.002, G1), "negative cell number in G1 $G1")
-    @assert(all(x -> x>=-0.002, G2), "negative cell number in G2 $G2")
+    @assert(all(x -> x>=0.0, G1[2:end]), "negative cell number in G1 $G1")
+    @assert(all(x -> x>=0.0, G2[2:end]), "negative cell number in G2 $G2")
     return G1[T] + G2[T]
 end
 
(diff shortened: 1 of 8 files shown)


aarmey:
Should be fixed now. finite_difference defaults to central differencing, which moves your current point up and down slightly. You can't go down if you're at 0. I switched it to forward differencing.

Farnazmdi:
Thank you very much.

Farnazmdi:
@aarmey should we merge this?

test/testHill.jl
```diff
@@ -12,4 +13,7 @@
 
     # Check that these at least run
     effects = getODEparams(params, conc)
+    eff2 = getODEparams(p2, conc)
+    num = numcells(effects[:, 3], g1[1]+g2[1], 100)
```
aarmey: Can you put this into a test set with a separate label? You’re specifically testing that this works with nd2 = 0, which is a different goal than the other testing code here.

src/Hill.jl
```diff
@@ -149,7 +152,7 @@ end
 function diffCell(params, g0, T)
     diffcells(x) = numcells(x, g0, T)
 
-    return Calculus.finite_difference(diffcells, params) / diffcells(params)
+    return Calculus.finite_difference(diffcells, params, :forward) / abs(diffcells(params))
```
aarmey: Don’t need abs() as values are always positive.

src/Hill.jl
```diff
@@ -138,9 +138,12 @@ end
 
 """ Calculate the # of cells in G1 for a set of parameters and T """
 function numcells(params, g0, T)
-    t = LinRange(0.0, 95.5, 192)
+    @assert(all(params .>= 0.0), "negative params $params")
+    t = LinRange(0.0, 200.0, 201)
     G1, G2 = predict(params, g0, t, Int(floor(params[6])), Int(floor(params[7])), Int(floor(params[8])), Int(floor(params[9])))
 
+    @assert(all(x -> x>=0.0, G1[2:end]), "negative cell number in G1 $G1")
```
aarmey: White vectorized assertion like params one above.

PR #171, opened 2020-02-29
LArray added for lower bound

Farnazmdi:
I added this to keep track of the name of the parameters. Did you mean this or having the labeled arrays somewhere else? @aarmey


src/allDrugs.jl
```diff
@@ -14,7 +14,7 @@ function getODEparamsAll(p::Array{Float64, 1}, concentrations::Array{Float64, 2}
         effects[4, :, i] = p[k + 5] .* xx
         k += 6
     end
-    effects[5, :, :] .= p[27]
+    effects[5, :, :] .= p[27] #percentage in G1
     effects[6, :, :] .= floor(p[28]) #nG1
```
aarmey: It'd be clearer to use the labelled names here, rather than indices, since you have them.

Farnazmdi: Sure. Will do.

Farnazmdi:
The problem is that, since I am running the tests and the notebooks with some estimated parameters, and I am not actually running the optimization, the type of the parameters are not `LArray`. So to change the code according to your comment I should change a whole bunch of code.
Basically anywhere that we want to use `getODEparamsAll` I have to have `LArray`, which is a little annoying.
Could I just have the labels as a comment?

aarmey:
Aha. Yes, if it's going to be difficult then don't bother—the benefit is only slight.

Farnazmdi:
Thanks.

PR #178, opened 2020-03-05
Combine all

Farnazmdi:


Farnazmdi:
Could you please take a look at the combination notebook in jenkins? @aarmey 
There are no more negative combination parameters, no more negative cell number, and made a test as you said in a new test file for combination.

Farnazmdi:
If they seem ready, I can send out the notebooks to Drs. Heiser and Gross.

combination.ipynb
```diff
@@ -0,0 +1,209 @@
+{
+ "cells": [
+  {
```
aarmey: Indicate time point of these plots. Some explanation of them at the top would be helpful, too.

src/Hill.jl
```diff
@@ -138,7 +138,7 @@ end
 
 """ Calculate the # of cells in G1 for a set of parameters and T """
 function numcells(params, g0, T)
-    @assert(all(params .>= 0.0), "negative params $params")
+#     @assert(all(params .>= 0.0), "negative params $params")
```
aarmey: Why comment this out?

combination.ipynb
Farnazmdi: Sure.

src/Hill.jl
Farnazmdi: Oops, sorry forgot to uncomment, although I have another assert in the BlissCombination that does the same.

Farnazmdi:
image

Farnazmdi:
Should I send the notebooks to Heiser lab?

PR #187, opened 2020-03-16
72 hours instead of 96 hours

Farnazmdi:
All set, only the combination parameters for doxorubicin and lapatinib have negative values, which is an assertion error in the `combination.ipynb`. I can let it run for more iterations (right now it ran for 30000 iters), to see if this problem will be solved.


Farnazmdi:
Should be all set now. There were no assertion errors.

PR #189, opened 2020-03-17
bliss for cell number

Farnazmdi:


src/allDrugs.jl
```diff
@@ -251,3 +251,23 @@ function combin2drugs(
...
+        combined[j,:,1] = -(num[:,1] .+ num[j,2] .- (num[:,1] .* num[j,2])) .+ (g1s[T,1,1] .+ g2s[T,1,1]) # lap w/ dox
+        combined[j,:,2] = -(num[:,1] .+ num[j,3] .- (num[:,1] .* num[j,3])) .+ (g1s[T,1,1] .+ g2s[T,1,1]) # lap w/ gem
+        combined[j,:,3] = -(num[:,1] .+ num[j,4] .- (num[:,1] .* num[j,4])) .+ (g1s[T,1,1] .+ g2s[T,1,1]) # lap w/ pac
+        combined[j,:,4] = -(num[:,2] .+ num[j,3] .- (num[:,2] .* num[j,3])) .+ (g1s[T,1,2] .+ g2s[T,1,2]) # dox w/ gem
+        combined[j,:,5] = -(num[:,2] .+ num[j,4] .- (num[:,2] .* num[j,4])) .+ (g1s[T,1,2] .+ g2s[T,1,2]) # dox w/ pac
+        combined[j,:,6] = -(num[:,3] .+ num[j,4] .- (num[:,3] .* num[j,4])) .+ (g1s[T,1,3] .+ g2s[T,1,3]) # gem w/ pac
+    end
+    @assert(all(combined .>= 0.0))
+    return combined
+end
```
Farnazmdi: 1. Subtracted the control cell # at time = 72hours.
2. for all 6 2-combinations for all 4 drugs, I wrote the bliss and added the control cell number back. But the assertion fails. The cell number goes negative

aarmey: Remember it needs to be a percent decrease. So, you divide by the control, then subtract off 1.0 (the control), then multiply by -1.0.

Farnazmdi: image
What I meant in the issue, was this:
The values for bliss combination are really large, I am not sure if they mean anything.

aarmey: Then something's wrong. You can't have more than a 100% decrease (so 1.0).

Farnazmdi: Did I have to do this for the parameters, too?

aarmey: Why is the variable here 3D? Shouldn't it be drug A by drug B?

Farnazmdi: I mean the percentage part.

aarmey: You're comparing this to your existing model using the parameters.

Farnazmdi: Sorry for the late response-- I went to get my monitor and other stuff from the office.
It is not 3D, the third dimension just identifies this is the combination of which two drugs.

Farnazmdi:
This is the one we had from the parameters:
image

This is Bliss with the cell number:
image

Farnazmdi:
Unfortunately, they don't really match.
For the case of parameters, I mean when I was doing the bliss for the parameters, did I have to divide them by their control value and then subtract 1 from them -- like here?
What I did there was I subtracted the parameter control from them and added them back at the end

Farnazmdi:
They are here: [link](https://transduc.seas.ucla.edu/job/meyer-lab/job/DrugResponseModel.jl/job/bliss/lastSuccessfulBuild/artifact/combination.html)

aarmey:
Seems like the single-drug conditions should match at least, since the model is fit to that exact data. Or am I missing something?

aarmey:
Maybe to help figure what's going on here, you can make a plot of the model prediction vs. data over concentration at just 96 hrs (on a separate branch)? It'll be helpful to see how well the model and single-drug data agrees.

Farnazmdi:
Sure, will do right away. It will clarify this to some extent.

Farnazmdi:
Sorry, I want to ask a question that I have asked before. Just to make sure; when we want to calculate the Bliss for cell number, we want them to be in fractional decrease; meaning, over the concentrations, if we look at the cell number, we want them to decrease. Is this right @aarmey?
What I have now is this:
I am dividing the cell number in each concentration by the cell number in control (which already makes them be in decreasing order over concentrations), then, I am subtracting them from 1.0
image
Here `i` refers to looping over all the 4 drugs, and `T` is the time point we want to calculate the cell number.
Could you please let me know if we need the *subtracting-1-part*?

Farnazmdi:
One more thing is that all of the drugs have a control condition trial for themselves. I know that we have set their parameters to be the same, but for example here, when I want to convert the units back by multiplying the control condition cell number, since it is a combination of two drugs, I don't know which control cell number to use.
If you agree, I could get rid of all of the control cell numbers except for one of them, and use that for all of the drugs -- in another branch.

aarmey:
Your subtracting 1 equation looks right to me.

aarmey:
Control is always the absence of any drug. Is the issue that the control conditions vary between drugs? I thought we resolved that by fitting all drugs at once?

Farnazmdi:
I don't think any issues come from that. We have the same parameters for them, but we still have 4 different control trials which are for our 4 drugs. I was wondering we remove 3 of them and just keep one. However, there is no problem arising from this yet. I said that because when I wanted to convert the cell numbers after applying Bliss back, I was confused to use which one. I just used the control condition for lapatinib for all of them.

Farnazmdi:
What I am wondering is that, are the two Bliss combinations we want to compare, comparable? I mean the approach we used for doing the Bliss combination for ODE parameters, were slightly different from the one we are doing for cell numbers. I mean, how much should we expect they give the same result?

aarmey:
Aren't they comparable in that they both provide us a prediction for the number of cells after treatment? If they disagree, that's interesting.

Farnazmdi:
Sure, I understand that. I just mean the way we (somehow) normalized the variable we want to apply the Bliss combination on, is different. Actually... their type is different. I guess I shouldn't worry about that.

Farnazmdi:
I double-check to see if I have made a mistake with the matrix indexes, otherwise, they disagree mostly.

Farnazmdi:
Hooray! It worked!!

Farnazmdi:
I mean the two agree.

aarmey:
Ok, is this ready to review then?

It makes sense that they usually agree, but hopefully they don't *always* agree. They are different notions of what drug additivity is, so if there are cases when they disagree, those are interesting examples to suggest to the Heiser lab. Of course, reality doesn't always match my hopes. :-)

Farnazmdi:
Tests are passing and notebooks are being built. I don't see what the problem is. I tried building it a couple of times.

Farnazmdi:
Yes, it is ready to review. There are some of them that don't quite match. But I was happy that I found my mistake in the code. It makes sense now.

PR #192, opened 2020-03-22
Temporal combination

Farnazmdi:
@aarmey This looks working. With this plotting function, we can choose which concentration of each to plot together.
Just one thing. Do you expect to see a smooth curve or jumps like this is alright?
image


src/ODEmodel.jl
```diff
@@ -69,7 +69,7 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2)
     end
 
     v = [ones(nG1) * p[5] * g_0 / nG1; ones(nG2) * (1.0 - p[5]) * g_0 / nG2; D1; D2]
-    A = ODEjac(p, t[2], nG1, nG2, nD1, nD2)
```
aarmey: How are you trying to adjust predict? With this change, it will only give you data with 30 min timesteps, which seems to *not* be what you want when you use it elsewhere.

src/ODEmodel.jl
```diff
@@ -69,7 +69,7 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2)
     end
 
     v = [ones(nG1) * p[5] * g_0 / nG1; ones(nG2) * (1.0 - p[5]) * g_0 / nG2; D1; D2]
```
aarmey: If you want to run the model with one drug then another, what you'll need to change is the initial vector after the switch. This is an initial value problem—you can't solve the ODE except by starting at t=0.

src/combination.jl
```diff
@@ -0,0 +1,171 @@
...
+    for j = 1:n
+        numscomb[j] = numcells(combination[:, j], g0, 96)
+        nums[j] = numcells(drugB[:, j], g0, 96)
+    end
+    plot(log.(concDrugB), numscomb, label = "pac + gemc", legendfontsize = 7, lw = 3, fg_legend = :transparent, shape = :circle, color = :purple)
+    plot!(log.(concDrugB), nums, label = "pac", lw = 3, xlabel = "log drug concentration", ylabel = "cell #", shape = :circle, color = :green)
+    plot!(dpi = 150)
+end
+
+function helperPlot(concd1, named1, concd2, named2, numscomb)
```
aarmey: Why is all this code moved? If you want to do it for cleanliness, please do it in a separate PR. The change you're suggesting for timed treatment will likely be complicated, so it will be helpful to have a clean diff.

Farnazmdi: Sure, I'll put it back and in a new branch will clean up.

src/ODEmodel.jl
Farnazmdi: Sorry, yes, I missed it. I Will work on it.

src/ODEmodel.jl
Farnazmdi: I will change this one, too. What I wanted to do was using the last data point from drug1 to be used as an initial condition in drug2.

src/combination.jl
Farnazmdi: Sorry, I will remove it and clean up in a new branch.

src/ODEmodel.jl
Farnazmdi: I thought it could start from any given time if we have the information at that exact time. In here, I am calling predict again, and giving it the last time point of the previous drug as initial values `g0`:
image

Isn't it taking care of this part you are mentioning?

aarmey: But the next predict can't just use the g1 and g2 quantities. It needs the entire v vector.

Farnazmdi: Sorry, I thought the `v` is being updated inside the predict here:
image

You mean I should input the v vector to the `predict`?

aarmey: Yeah, you'll need to output this and be able to provide it to a predict function. It's the actual starting point for the ODE system. G1 and G2 are summary quantities.

aarmey: Why don't you write a new, separate version of predict() that just takes a single time point and gives you back the full species vector, using Krylov solving (`ExponentialUtilities.jl`)? It's imperative that you understand what this function is doing.

Farnazmdi: Thank youuu, I will do that.

src/ODEmodel.jl
```diff
@@ -77,13 +77,31 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2)
...
+        D2 = Float64[]
+    else
+        D2 = zeros(nD2)
+    end
+
+    dt = t[2] - t[1]
+    v = [ones(nG1) * p[5] * g0 / nG1; ones(nG2) * (1.0 - p[5]) * g0 / nG2; D1; D2]
+    A = ODEjac(p, dt, nG1, nG2, nD1, nD2)
+    return vector
+end
```
Farnazmdi: Here is the new predict version. It returns `v` for a given initial condition.

src/allDrugs.jl
```diff
@@ -273,3 +273,46 @@ function blissCellNum(g1s, g2s; T = 96, n = 8)
...
+    nD1 = Int(floor(params2[8]))
+    nD2 = Int(floor(params2[9]))
+    v = predict2(params2, g1L[end] + g2L[end], t2, nG1, nG2, nD1, nD2,2)
+    g1G[1] = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
+    g2G[1] = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
+    for j=2:96
+        v = predict2(params2, g1G[j-1]+g2G[j-1], t2, nG1, nG2, nD1, nD2,j+1)
+        g1G[j] = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
+        g2G[j] = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
+    end
```
Farnazmdi: And here, I'm using the `v` vector that `precidt2` is producing, and calculating each point of G1 and G2 using that vector.
The problem is that it is returning a very large or very small number. I checked it with the first value that we get in `predict`, and `v` was correct. The `g0` I'm providing it every time is very large though.

Farnazmdi: Sorry, I had missed a line.
It works I guess:
image

Farnazmdi: This is when we use first 48 hours with control, and the next 48 hours with concentration number 7 of gemcitabine
image

src/allDrugs.jl
```diff
@@ -273,3 +273,46 @@ function blissCellNum(g1s, g2s; T = 96, n = 8)
...
     return combined
 end
+
+""" temporal combination of two drugs.
+Here we want to plot a simulation, with first 48 hours being treated with lapatinib, and another 48 hours with gem. """
+function temporal_combination(params1, params2, g0, title)
+    t1 = LinRange(0.0, 48.0, 96)
+    t2 = LinRange(48.5, 95.5, 96)
+
+    g1L, g2L = predict(params1, g0, t1, Int(floor(params1[6])), Int(floor(params1[7])), Int(floor(params1[8])), Int(floor(params1[9])))
```
aarmey: But you need the state vector from here.

src/allDrugs.jl
```diff
@@ -273,3 +273,46 @@ function blissCellNum(g1s, g2s; T = 96, n = 8)
...
+    g1G = g2G = zeros(96)
+    nG1 = Int(floor(params2[6]))
+    nG2 = Int(floor(params2[7]))
+    nD1 = Int(floor(params2[8]))
+    nD2 = Int(floor(params2[9]))
+    v = predict2(params2, g1L[end] + g2L[end], t2, nG1, nG2, nD1, nD2,2)
```
aarmey: `g1L[end] + g2L[end]` is not the correct state at t=48.0. This assumes cells are evenly distributed through the linear chain, which isn't the case as soon as you treat with a drug.

aarmey:
I'll write an alternative predict implementation.

PR #196, opened 2020-03-23
Use Krylov solver if we just need one time point.

aarmey:
In a lot of places, you reference the solution at time `t` by the index `T` within a LinRange. This change lets you just pass time `t` to predict, which should be much, much faster.


Farnazmdi:
Thank you.
I guess the tests also use `predict`, they need to be edited too.

Farnazmdi:
Oh no sorry, never mind. I thought t was removed from predict. Sorry.

Farnazmdi:
I'll do the rest in a new branch.
Thank you.

PR #197, opened 2020-03-24
temporal combination

Farnazmdi:


src/ODEmodel.jl
```diff
@@ -62,16 +62,17 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2, retVec =
...
+        G1 = sum(vec[1:nG1]) + sum(vec[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
+        G2 = sum(vec[(nG1 + 1):(nG1 + nG2)]) + sum(vec[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
+        newV = expv(t, A, vec)
+        return G1, G2, newV
     else
         # Some assumptions
         @assert t[1] == 0.0
+        v = [ones(nG1) * p[5] * g_0 / nG1; ones(nG2) * (1.0 - p[5]) * g_0 / nG2; D1; D2]
         rmul!(A, t[2])
         A = LinearAlgebra.exp!(A)
```
Farnazmdi: I have been struggling with this for quite a few hours. For some parameter sets, the state vector values go to almost infinity. Do you see anything wrong with these changes @aarmey?
I inputted the `v` vector to the `predict`. I moved the `expv(...)` to the bottom because the state vector it receives is actually `v = A * v`.

Farnazmdi: The time we give the `expv()` has to be `dt` or just `t`?

Farnazmdi: I tried both, didn't change the infinity result, though.

aarmey: The way you've changed the function, you have to be careful to set `vec` to something if `t` is a time value. How about have `vec = nothing`, then if `vec is nothing` set it equal to the default value for `v`?

aarmey: After this, could you provide a set of function arguments that cause this to be infinite? This inside a test would be very helpful.

src/ODEmodel.jl
```diff
@@ -62,16 +62,17 @@ function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2, retVec =
...
 
-    v = [ones(nG1) * p[5] * g_0 / nG1; ones(nG2) * (1.0 - p[5]) * g_0 / nG2; D1; D2]
     A = ODEjac(p, nG1, nG2, nD1, nD2)
 
     if t isa Real
-        v = ExponentialUtilities.expv(t, A, v)
-        G1 = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
-        G2 = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
+        G1 = sum(vec[1:nG1]) + sum(vec[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
+        G2 = sum(vec[(nG1 + 1):(nG1 + nG2)]) + sum(vec[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
```
aarmey: Why are you taking the sum of G1 and G2 before `expv`? This is just the starting vector you passed in.

Farnazmdi: The vector it is taking comes from the output of the `else` part, which is already A*v. I thought if I put the `expv` above, I would double exponentiate it.

src/ODEmodel.jl
Farnazmdi: I will try it right now.

src/ODEmodel.jl
aarmey: What is the vector being passed in here? The way to think about this is not where it has been exponentiated, but what it represents.

Farnazmdi: This is how I thought about it:
First, we call predict for the first 48 hours for lapatinib with a time vector and return the state vector at the last time point.
Then we pass in the state vector so that the `if statement` is true, so it will produce the first data point after 48 hours. We loop through this and successively pass the produced state vector to build the next data point and state vector.

aarmey: Ahh... you don't ever need to use the `ExponentialUtilities.expv` part of this function for the temporal combination. The **only** reason I added this is for when you only need to calculate one time-point. The **only** change you need for temporal combinations is to be able to pass in the starting state vector. The trick is, if you're switching drugs, t=48hrs is effectively t=0 as far as predict is concerned.

src/allDrugs.jl
```diff
@@ -273,3 +273,19 @@ function blissCellNum(g1s, g2s; T = 96, n = 8)
     @assert(all(combined .>= 0.0))
     return combined
 end
+
+""" Function for calculating temporal combination of two drugs. """
+function temporal_combination(params1, params2, g0)
+    t1 = LinRange(0.0, 47.5, 96)
+    t2 = LinRange(48.0, 95.5, 96)
```
aarmey: i.e. t2 here should start at 0.0 and go to 95.5-48.0. That's also likely why you're getting really large numbers.

src/ODEmodel.jl
Farnazmdi: Oh, I see. Thank you. I'll make the changes right away.

Farnazmdi:
@aarmey thank you very much. I was confusing myself all day!
The curves are smooth and there is no infinity. I will make a plot function in a new branch if you agree.

aarmey:
Great!

src/ODEmodel.jl
```diff
@@ -50,7 +50,7 @@ end
 
 
 """ Predicts the model given a set of parametrs. """
-function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2, retVec = false)
+function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2, vec)
```
aarmey:
```suggestion
function predict(p, g_0::Real, t, nG1::Integer, nG2::Integer, nD1, nD2, vec = nothing)
```

This will make `vec` default to nothing unless you provide it.

Farnazmdi: Thanks

Farnazmdi:
I guess the server is very busy right now. I'll try to build again in half an hour.

Farnazmdi:
@aarmey is it alright to just remove the argument type for `g0`?

aarmey:
Sure.

PR #200, opened 2020-03-24
plot temporal combination

Farnazmdi:


Farnazmdi:
Oh, I'm gonna need to unify the time-axis maximum.

Farnazmdi:
Should I send all of the combinations to Drs. Gross and Heiser? Or this is enough? I thought I could ask them to see which combinations they intend to see, and then I can simulate provide to them.

Farnazmdi:
image

aarmey:
Are any drugs/concentrations really different in the final cell # during the A->B direction compared to the B->A direction?

Farnazmdi:
A little yes, I will send you a file I have prepared that has all the 6 combinations, all in an intermediate concentration.

PR #202, opened 2020-03-24
Clean up

Farnazmdi:


Farnazmdi:
The `allDrugs.jl` file was very full, so I just moved the functions related to any sort of combination to a new .jl file.

PR #204, opened 2020-03-25
Weave

Farnazmdi:
I followed the tutorials, but not sure it is the right way. Could you please help me with how to include the proper command in the `makefile`?


src/notebookPlots.jmd
```diff
@@ -0,0 +1,164 @@
...
+plot(pl...)
+plot!(size=(1200, 800))
+```
+
+```{julia; echo=false; skip="notebook"}
+tmp = joinpath(DrugresponseModel.jl,src,notebookPlots.jmd)
+dir = jointpath(DrugResponseModel.jl)
+args[:doctype] = "html"
+weave(tmp,doctype = "md2html",out_path=dir,args=args; fig_ext=".svg")
+```
```
Farnazmdi: This is the part that I called the `weave`. @aarmey

aarmey:
Great. I'll take this PR from here. Please branch off for other changes.

Farnazmdi:
Thank you.

PR #208, opened 2020-03-26
plot time series for IC50 conc of drugs

Farnazmdi:


Farnazmdi:
image
This is what I have for drugs at their IC50.

Farnazmdi:
Please look at it as a 2 row and 3 columns. Each plot actually is 4 sub-plots. When providing to Heiser lab, I will draw a line to separate them.
Do these results make sense to you?

aarmey:
These make sense to me.

Farnazmdi:
I thought I add the difference of Bliss combination in another branch, but since you haven't reviewed this, I will add it in this branch, if you agree.

PR #210, opened 2020-03-26
some edits and 3 plots together

Farnazmdi:


Farnazmdi:
The third plot on the right is the difference between the two left plots.
image

PR #216, opened 2020-04-01
Explain

Farnazmdi:
More explanations added to the Combination notebook.


Farnazmdi:
@aarmey I added some explanation, but 2 files are not being built. I didn't change anything in `allDrug.jmd`
I found this 
image
Could you please let me know what I should do if you have encountered this before?

aarmey:
You can see more output if you run `make` outside of Jenkins. Just make a change... let me know if it fixes things.

Farnazmdi:
Plotnotebooks.pdf is very messy, nothing is in its place. I guess it might have something to do with one of the automatic PRs that updates the .toml file. I approved it, it is #214. I'm not sure if it is, but these new versions might not be compatible as the previous versions.

notebookPlots.jmd
```diff
@@ -18,137 +13,166 @@ effects = getODEparams(ps, conc_l)
 g0 = g1[1] + g2[1]
 T = 170
 ```
+
+### Lapatinib
+
 #### Plotting the gradient
+
```
aarmey, line 20: It looks like you might need to have some non-header text here to ensure the figure is left-justified.

notebookPlots.jmd
```diff
@@ -18,137 +13,166 @@ effects = getODEparams(ps, conc_l)
...
 ```julia; echo=false
 plotGradient(effects, conc_l, g0, T)
 ```
+
 #### Plotting the time-series simulations compared to the actual data.
```
aarmey, line 25: Oh, also, `####` is a paragraph header, where content should come to the right of it, while `###` is a subsubheading, so all content will show below. I bet if you change all the `####` to `###` and all the `###` to `##`, that'll fix the problem.

Farnazmdi:
That's really weird, I'm sure I pushed a commit almost an hour ago and saw the in the Jenkins that all were correct, except for notebookPlots.pdf, but I don't see any trace of that push!!

Farnazmdi:
I still can't generate the combination.jmd (now, I have it as 2 separate files: Bliss.jmd and temporal_combination.jmd)

Could this be our problem? @aarmey 

image

aarmey:
Do they work with `make all` locally? Or is this a problem both locally and on Jenkins?

Farnazmdi:
It's both local and Jenkins problem.

Farnazmdi:
In my local it creates a .tex file and asks me to create the PDF manually! I tried using overleaf and it was something like a template, without the figures. I couldn't find the folder it was referring to for the figures.

aarmey:
Is local Aretha or your computer?

Farnazmdi:
Yes.

Farnazmdi:
I mean Aretha.

aarmey:
Running `xelatex Bliss.tex`, I get the error:

```
! Undefined control sequence.
<argument> \endash 
                   
l.134 ...e cell number values \ensuremath{\endash}
                                                   instead of drug effect pa...
```

The easiest fix would be to not use en-dashes (`--`).

Farnazmdi:
Thank you so much.

PR #236, opened 2020-04-12
Replicates

Farnazmdi:


Farnazmdi:
@aarmey should all the plots -- including the time-series simulations -- have the ribbon plot or just the estimated parameters vs. concentration?

aarmey:
Wouldn’t the time series plots be the easiest to add first?

Farnazmdi:
Actually, I guess they are harder than just the parameters.

Farnazmdi:
I thought of it like this:
I have to find the time-series solution for the 3 replicates, and find the mean and standard deviation of the three sets.
But for the parameters, I don't need to use `predict` and find cell numbers. I have the parameters, so I just find the mean and standard deviation.

aarmey:
Oh, but I don’t mean showing the model uncertainty. I mean just plotting the data with error.

Farnazmdi:
The representation would also be a little messy if I want to show G1 cell#, G2 cell#, and total cell# and the model uncertainty.

Oh, yes I can do that first. I ran the `allDrugs` for 60K iterations and I have the estimated parameters for the 3 replicates. I can do those plots, too.

aarmey:
Sounds good.

Farnazmdi:
image

Farnazmdi:
Lapatinib parameters:
image

Doxorubicin parameters:
image

Gemcitabine parameters:
image

Paclitaxel parameters:
image

Palbociclib parameters:
image


Looks like G2 prog. rate is quite flexible. I guess it's good news because we had the most difficulty interpreting changes in this parameter. And bad news, maybe we should break it into more parameters to observe a meaningful behavior?

aarmey:
We could use the single cell measurements to constrain the control progression rates. This might require reformulating the progression rates to be (rate) * (# steps) so that the constraint is independent of the number of steps.

Farnazmdi:
Sorry, could you please elaborate on how to use the single-cell measurements to constrain the model?

aarmey:
From the single cell measurements, you have the distribution of G1 and G2 lifetimes. Those tell you something about G1/G2 progression rates.

Farnazmdi:
Got it. Thanks.
How about, I set the lower and upper bound of the progression rate based on the distribution parameters? Did you mean this by constraining?

aarmey:
Exactly.

Farnazmdi:
I added a .jmd file for the replicates and their simulations. The last series of plots are the "mean of simulation with std of data". I wasn't sure which configuration would be best, so I just plotted the average of simulation results, along with the std of the original data. Fortunately, the model falls into the bounds of the data, for all drugs. I just included the lapatinib here, but it is the matter of changing the counter parameter to show another drug.
I guess this is really heavy adds and deletions. If you agree, we merge this and I address the rest of the issues in another branch. Most of the lines are just the addition and deletion of the data files.

Farnazmdi:
I looked at all of them. They all fall into the bounds pretty well!

Issue #240, opened 2020-04-14
Use single cell data to constrain the prog rates


Farnazmdi:

Farnazmdi:
I guess with this, and smoother raw data we might get better fits even!

Farnazmdi:
This is what we have for now. I assume we want to keep the control the same in all of the trials, and have only one shape parameter across the drugs. I will do this now.
![image](https://user-images.githubusercontent.com/34812750/79418773-476fed80-7f6a-11ea-9c55-a263551d48d5.png)

Farnazmdi:
![image](https://user-images.githubusercontent.com/34812750/79473299-c9d3ce00-7fb9-11ea-9e25-8cfc22f4ac72.png)

PR #245, opened 2020-04-16
adding constrain

Farnazmdi:
Realized that if we simulate all the 3 drugs I had single-cell data for, this would be the result of the shape and scale parameters:
#240 
So I changed the bounds accordingly and ran the model for allDrugs again,
The result is in replicates.pdf.


Farnazmdi:
It looks like these changes have not improved the std of G2 progression rate.

PR #249, opened 2020-04-16
optimize with local optim

Farnazmdi:
It takes forever to run. I set the iteration number to be at 5.


src/allDrugs.jl
```diff
@@ -214,3 +214,51 @@ function avgRepsParams(rep1, rep2, rep3, concs)
...
+        3,
+        15,
+        0,
+        0,
+    )
+    highPiece = [10.0, 7.0, 2.3, 3.0, 3.0]
+    high = vcat(
+        maximum(concs[:, 1]),
+        highPiece,
+        maximum(concs[:, 2]),
```
Farnazmdi, line 244: Is this function what you meant?
Should I keep the notebook?

Project.toml
```diff
@@ -8,9 +8,13 @@ BlackBoxOptim = "a134a8b2-14d6-55f6-9291-3336d3ab0209"
 CSV = "336ed68f-0bac-5ca0-87d4-7b16caf5d00b"
 Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
+DiffEqParamEstim = "1130ab10-4a5a-5621-a13d-e4788d82bd4c"
```
aarmey, line 11: You only need the Optim package here.

optim_all.ipynb
```diff
@@ -0,0 +1,78 @@
...
+     "output_type": "stream",
+     "text": [
+      "┌ Info: Precompiling DrugResponseModel [f46a7e40-ab32-11e9-0ff4-0323d592fbfd]\n",
+      "└ @ Base loading.jl:1260\n"
+     ]
+    },
+    {
+     "data": {
+      "text/plain": [
+       "37-element Array{Float64,1}:\n",
```
aarmey, line 19: Can you get rid of the notebook but call it within the tests?

src/allDrugs.jl
aarmey: Yep, this looks good.

Project.toml
Farnazmdi: Yes, sorry, I was looking for the package that has LevenbergLarquart. Ended up not using it.

Farnazmdi:
Could you please approve if it is ready @aarmey ?

aarmey:
This all looks good, but I'm going to explore some possibilities on this branch. I'll tag you when I have a chance to try a few adjustments.

Farnazmdi:
Should I use this optimization for estimating the parameters?

aarmey:
You could use it to "polish" after bboptim, but we'd need to handle the gradients for the step number parameters before only using this.

Farnazmdi:
I will merge it and then see how it performs in another branch because Julia notebook in this directory still doesn't work, but other places it looks working.

PR #251, opened 2020-04-17
combination measure

Farnazmdi:
replaced Bliss.pdf with combination.pdf


Project.toml
```diff
@@ -11,6 +11,7 @@ DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Measures = "442fdcdd-2543-5da2-b0f3-8c86c306513e"
+Plotly = "58dd65bb-95f3-509e-9936-c39a10fdeae7"
```
aarmey, line 14: Does it not work with the GR backend?

Farnazmdi: I guess the `clim` that defines the heat map color limits needs it. The example I saw was with plotly.

Farnazmdi: [here](https://github.com/JuliaPlots/Plots.jl/issues/293)

aarmey: Aha. Indeed, looks like it's only in Plotly right now.

Farnazmdi: It is only for one replicate. Do we want to have an average of the three or this suffices for now?

aarmey: I think ultimately you want to build the plot for each replicate, then average those.

Issue #253, opened 2020-04-17
Let's just remove this gradient idea


aarmey:
It seems to be more confusing than it's worth.

https://github.com/meyer-lab/DrugResponseModel.jl/blob/91b81806a2735a6837572e2b57b3ebb77d0e4509/src/Hill.jl#L132

Issue #252, opened 2020-04-17
Why does the data not match in the model fit vs data plot?


aarmey:
The data plotted in replicates.pdf is different in figure 1 and 2, even though it should be the same I think...

aarmey:
Also the line of data average will need to be removed, since the line represents the model when the two are overlaid.

PR #254, opened 2020-04-17
Issues

Farnazmdi:
#252 
#253 
Addresses these two issues.


Farnazmdi:
The model then is not as good as what we thought it is, but I am afraid in the last PRs I had forgotten to replace the parameters with the new estimated ones, that's why we didn't see any change in the std of G2 prog. rate. The good news is the parameters have a narrow uncertainty and pretty small std.

PR #261, opened 2020-04-21
Perform differencing for integer-valued parameters.

### LINKED ISSUE #104: Custom differencing scheme for Jacobian
Should enable local optimization despite integer values. Can discuss.

aarmey:
Fixes #104 

This now properly handles the integer-valued parameters. To the extent that the optimization problem is convex, this should work as well or better than bboptim.

### DIFF
--- a/src/allDrugs.jl
+++ b/src/allDrugs.jl
@@ -222,9 +222,20 @@ end
 
 function optim_all(concs::Array{Float64, 2}, g1::Array{Float64, 3}, g2::Array{Float64, 3})
     f(hillParams) = residHillAll(hillParams, concs, g1, g2)
-    g!(G, hillParams) = ForwardDiff.gradient!(G, f, hillParams)
 
-    low = vcat(ones(32) * 1.0e-9,  0.4, 3, 15, 0, 0)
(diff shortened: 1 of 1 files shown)


src/allDrugs.jl
```diff
@@ -222,9 +222,20 @@ end
...
+        ForwardDiff.gradient!(G, f, hillParams)
+        costCenter = f(hillParams)
+
+        # Handle the integer-valued parameters
+        for ii = 34:37
+            pp = copy(hillParams)
+            pp[ii] += 1.0
+            G[ii] = f(pp) - costCenter
+        end
+    end
```
Farnazmdi, line 236: Could you please give me a clue of how this is handling the integer-valued parameters? What is `G` here?

aarmey: G is the gradient, or `d[cost]/d[param_i]`. The issue is that autodiff correctly determines that this is zero for the integer-valued parameters because they are discontinuous. So, instead, what I do is increase that parameter value by `1.0`, so that it reaches the next integer. I then take the difference to estimate the slope. This is essentially forward differencing, but with a big step size.

PR #264, opened 2020-04-21
edits to separately fitted drugs

Farnazmdi:


Farnazmdi:
Should be ready to merge. Without a notebook it takes a lot of time debugging!
Gemcitabine and palbociclib also look good.

aarmey:
Are the notebooks still not working, or is it because the documents are in Weave.jl? If the latter, note that you can convert a Weave doc to a jupyter notebook and back:

http://weavejl.mpastell.com/stable/notebooks/#Weave.convert_doc-Tuple{String,String}

Farnazmdi:
It is the problem I had yesterday. Julia kernel is unknown.

PR #267, opened 2020-04-22
clean up

Farnazmdi:
I used the functions I wrote for replicates of individual drugs for "AllDrugs".
Also, I think I have to go back to the previous estimations before limiting the nG1 and nG2 for estimating all drugs at once. The plot for G1 and G2 prog. rates and death rates don't look right in any of the drugs. 
Do you agree @aarmey ?

Also, I polished the estimates with `optim_all`, and the results you see in the AllDrugs notebook is after going through the optim_all. It doesn't seem improved very much, but I tried it for 30 iterations. I don't have a sense of how many iterations would suffice.


Farnazmdi:
One thing I noticed is that the EC50 for drugs is being estimated very differently in individual estimation and at once.

aarmey:
"Also, I think I have to go back to the previous estimations before limiting the nG1 and nG2 for estimating all drugs at once."

What do you mean by this?

aarmey:
For optim_all, you should remove the iterations limit so it runs until it exits because of convergence.

src/DrugResponseModel.jl
```diff
@@ -56,6 +56,10 @@ export setup_data,
     savitzky_golay_filter,
     optim_all,
     heatmap_combination,
-    plot2
+    plot2,
```
aarmey, line 59: You don't need to export every function. In fact, I think it might be better to not export _any_ function, then call them from the Weave files using `DrugResponseModel.function()`.

Farnazmdi:
What I mean by that sentence is that, recently, I restricted the lower and upper bounds for nG1 and nG2 based on the single-cell data. Before that, the dose-response curves of the ODE parameters were more similar to the ones we see when we fit the drugs individually.

src/DrugResponseModel.jl
Farnazmdi: Sure.

aarmey:
I think it's fine to widen the constraint.

Farnazmdi:
I simulated the drugs at once again, only gemcitabine looks very good, although the cost in optimization was decreased by almost 1000, and I did the `optim_all` after `bboptim` was done. :(
The progression rates are still off but I did some cleaning in this branch I guess it worth merging.

Farnazmdi:
They might haven't changed, but because we look at them with the error bar it looks worse.
If you mean the first time I plotted them with error bar, it was my mistake, because it was automatically plotting the standard deviation based on std of the *simulations* and not the one from data that I was passing to it.

Issue #270, opened 2020-04-23
compare temporal combinations


Farnazmdi:
Write a function to calculate the difference between an awful lot of time-combinations and show which one is different with the reference.

Farnazmdi:
Looks like any combination with taxol, the order of treatment matters.

Farnazmdi:
Have an idea that could potentially be used for sequential treatment, I guess.

What if we consider each drug treatment a state, and the cell number after that drug treatment the observation. In the framework of the tHMM, could we use this to decide which drugs could potentially be scheduled one after another? I understand the "for how long" would be a little challenging, but in the current setting we have the duration of time and the number of generations that the states persist.
Do you think this is a reasonable approach @aarmey ?

aarmey:
What is the tree then?

Farnazmdi:
My mistake, sorry. Just the HMM.

aarmey:
I'm not sure I see the value over running the model we have.

Farnazmdi:
First, I was thinking in terms of a tHMM with including single-cell fates as another observation and the time duration that a state exists until it switches to another. But I guess with the drugs being different states, then there shouldn't be any tree structure. Just the idea of state switching, I guess.
And that Single-drug effects could be used from the ODE model to give the state distributions. 
Does it make sense at all?

PR #273, opened 2020-04-26
Compare

Farnazmdi:


Farnazmdi:
#270

Farnazmdi:
I will work on it after the qual, if that's alright. I have to correct some part so that the time-series simulations in the temporal combination work.

aarmey:
Sounds good.

Farnazmdi:
This PR does  the following:
for any set of (drugA_conc_x, drugB_conc_y) finds the different simulation results in terms of cell number for the two different orders of treatment with drugA and drugB, and outputs a heatmap showing the difference, and plots the time-series of the one with the highest difference between the two treatment orders

Farnazmdi:
The results are in `temporal_combin.pdf`.
I will take care of the rest in another PR. 
Should I merge this @aarmey ?

Farnazmdi:
Since you've approved it before, I guess it is fine to merge.

PR #283, opened 2020-05-18
Fix at-once

Farnazmdi:
After spending hours I couldn't figure out what the problem was, but I tried different previous versions of the code and one of them worked with acceptable speed.
I would replicate only very few lines in the `residHillAll()`, and I returned the `ODEjac()` and `predict()` to their previous forms. Sorry, I know this is not the best strategy, but it works. 
I did a little bit of clean up, too.
Also, I believe the Bliss combination I have written is not completely correct. If you agree to merge this I can address that in another PR.
Would it be alright to merge this @aarmey ?


Farnazmdi:
I also compared the simulations and estimated parameters from "individual-drug simulations" with "all-at-once simulations" and they match.

aarmey:
Do you know the problem goes away with putting back the old ODE model, or could it be from the other changes in the code here? In other words, if you keep all the changes here *except* the changes to predict, is it still broken?

src/allDrugs.jl
```diff
@@ -120,7 +146,7 @@ function optim_all(concs::Array{Float64, 2}, g1::Array{Float64, 3}, g2::Array{Fl
         costCenter = f(hillParams)
 
         # Handle the integer-valued parameters
-        for ii = 34:37
+        for ii = 33:37
```
aarmey, line 149: E.g., if the problem is predict(), why do you need this change?

Farnazmdi: Sorry, this was just my mistake. I change a lot of things and when bringing back the code I must have made a mistake. This shouldn't change at all. I will correct it right away.

Farnazmdi:
I will try that and update you here.

Farnazmdi:
I did try to only replace `predict()` and I had to use matrix exponential methods for the sparse matrix. It looks like it just gets stuck somewhere, even when I don't run the optimization and just call the `predict()` it gets stuck. I couldn't spot where yet, or it might be that my internet connection is very slow that it doesn't run (maybe).
Unfortunately, since yesterday our WiFi has been broken. I use the hotspot of my phone but it is slow.

src/ODEmodel.jl
```diff
@@ -44,80 +44,72 @@ end
...
-        G2 = Vector{T}(undef, length(t))
+        G1 = sum(v[1:nG1]) + sum(v[(nG1 + nG2 + 1):(nG1 + nG2 + nD1)])
+        G2 = sum(v[(nG1 + 1):(nG1 + nG2)]) + sum(v[(nG1 + nG2 + nD1 + 1):(nG1 + nG2 + nD1 + nD2)])
     else
-        cost = 0.0
-    end
+        # Some assumptions
+        @assert t[1] == 0.0
+        rmul!(A, t[2])
+        A = SparseMatrixExp(A)
```
Farnazmdi, line 82: Could you please let me know if these two functions for the sparse matrix are correctly here or not? It might be that these two are very slow or just loop forever.

aarmey: I'm not sure what you're trying to do? The exponential of a sparse matrix is typically not sparse, so I'm not sure how this function would exist. You could use Krylov solving with a sparse matrix, though. In that case you'd be calling something like expmv at each time point.

aarmey:
The matrix doesn't have to be sparse. Use diagm to give you a dense matrix, and then you can use the standard exponential methods.

Farnazmdi:
Yes it was the matter of variable type that I was getting an error in exponentiation of the matrix, will do, thanks.

Farnazmdi:
I checked the optimization, too. It works well.

aarmey:
Still have the same question... was the problem in `predict()` or elsewhere?

Farnazmdi:
Yes, it was `predict()`, but I'm not sure which line exactly. I only replaced `predict` with a previous version, and returned the Jacobian matrix type to a dense matrix instead of sparse.
One thing that doesn't work now, is the local optimization, which errors because `exp!()` cannot work on `DorwardDiff.Dual(...)`. I searched to find a fix for it but didn't find a solution. I guess something in the predict was taking care of that, and by returning it back to a previous version, the local optimization is broken now.

aarmey:
Ok. I can fix that Dual type error later in another PR.

replicatesAtOnce.jmd
```diff
@@ -27,7 +42,7 @@ G2_1 = zeros(189, 8, 5)
 G2_2 = zeros(189, 8, 5)
 G2_3 = zeros(189, 8, 5)
 for i=1:5
-    G1_1[:, :, i], G2_1[:, :, i], G1_2[:, :, i], G2_2[:, :, i], G1_3[:, :, i], G2_3[:, :, i] = DrugResponseModel.predict_replicates(ef1[:, :, i], ef2[:, :, i], ef3[:, :, i], concs[:,1], g0)
```
aarmey, line 30: Why this change?

Farnazmdi:
Do you have a specific command line syntax for catching the unused packages? I didn't do it because I thought there are dependencies that I might not know and delete.

aarmey:
Re unused packages: Nope. Purely just guessing here. We'll see if the build finishes successfully.

replicatesAtOnce.jmd
Farnazmdi: I had two functions to plot the three replicates along with the simulations; one for the time we were simulating them individually (`plot2()`), and one for the one simulating them at once (`plot_ribbon_replicates()`--I guess). I just deleted one of them to avoid repeated code.

Farnazmdi:
Thanks.

PR #286, opened 2020-05-21
Loewe

Farnazmdi:


Farnazmdi:
@aarmey I can't figure out what's wrong that it can't find a real root for this. Would it be possible for you to take a look at it, please?
I will explain what I have done in a second.

src/combination.jl
```diff
@@ -252,3 +250,32 @@ function plot_order_temporalCombin(params1, params2, g1s, g2s, named1, named2)
...
 end
+
+####--------------- Loewe additivity ------------------####
+""" Find the inverse of a hill function (concentration), given the parameters and the effect. """
+function inv_hill(p::Array{Float64, 1}, y)
+    #p = [EC50, min, max, steepness], y:effect. it returns concentration
+    conc = p[1] * (((y - p[2]) / (p[3] - y)) ^ p[4])
+    return conc
+end
+
```
Farnazmdi, line 261: Here is just the Hill inverse function.

src/combination.jl
```diff
@@ -252,3 +250,32 @@ function plot_order_temporalCombin(params1, params2, g1s, g2s, named1, named2)
...
+""" The combination Index for Loewe. """
+function loewe(d1, p1, d2, p2, conc1, conc2)
+    # x: the combined effect
+    f(x) = (d1 / inv_hill(p1, x)) + (d2 / inv_hill(p2, x)) - 1.0
+
+    combined_effect = find_zero(f, [(minimum([p1[2], p2[2]]) + eps()), (maximum([p1[3], p2[3]])-eps())])
+    return combined_effect
+end
```
Farnazmdi, line 269: Here I wrote the combination index for the Loewe. 
The lower and upper bound to find the roots are set to be between the min and max parameters, due to the kth root in the inverse Hill function. and the epsilons are just to avoid being equal to min and max since it creates zeros in the denominator.

src/combination.jl
```diff
@@ -252,3 +250,32 @@ function plot_order_temporalCombin(params1, params2, g1s, g2s, named1, named2)
...
+function apply_loewe(params1, params2, conc1, conc2)
+
+    combined_effects = zeros(length(conc1), length(conc2))
+    for (ind1, d1) in enumerate(conc1)
+        for (ind2, d2) in enumerate(conc2) # keep d1 constant and loop over d2
+            combined_effects[ind1, ind2] = loewe(d1, params1, d2, params2, conc1, conc2)
+        end
+    end
+    return combined_effects
+end
```
Farnazmdi, line 281: Please ignore this for now, it just loops over all the concentrations to find all the combinations.

Farnazmdi:
I tried it in a notebook, named untitled. I can't think of anything that is wrong here, but it can't find a root and errors.
image

src/combination.jl
```diff
@@ -252,3 +250,32 @@ function plot_order_temporalCombin(params1, params2, g1s, g2s, named1, named2)
         title = string(named1, " --> ", named2, " - ", named2, " --> ", named1),
     )
 end
+
+####--------------- Loewe additivity ------------------####
+""" Find the inverse of a hill function (concentration), given the parameters and the effect. """
+function inv_hill(p::Array{Float64, 1}, y)
+    #p = [EC50, min, max, steepness], y:effect. it returns concentration
+    conc = p[1] * (((y - p[2]) / (p[3] - y)) ^ p[4])
```
aarmey, line 258: You're getting an error because the term being exponentiated is negative, which ends up creating a complex number. `y` can never be below `p[2]`, because that's the minimum of the function. Otherwise you're asking for the inverse of a value outside of the function's range, which isn't defined.

The minimum of both Hill curves might have to be shared because of this... it might help to see whether this is the case in other applications of Loewe.

aarmey: (p.s. Eid Mubarak!)

Farnazmdi: Thank you, I will look into other papers that used Loewe.

Farnazmdi: Thanks :)

Farnazmdi: Oh yes, I remembered that Loewe is used in the case when the two drugs have the same minimum and maximum, and there is just a shift along the concentration axis. But I thought it is just theoretical! I still have to look at other papers.

Farnazmdi:
Edit to the boundary that we expect the roots to fall into:
image
However, it doesn't work for "all" of the concentrations.

Farnazmdi:
image

image

I just set any of the non-real solutions to be zero using a `try and catch`.
Please let me know if it makes sense when you get a chance. Thanks

aarmey:
Why would Loewe additivity be for the rates, and not the cell # itself?

Farnazmdi:
I was trying to replace it with Bliss, so we could use the model predictions as a representative of the experiments and for that we needed to have the expected parameters in the combination. 
Should I just apply that for the cell number?

Farnazmdi:
You mean I find the best Hill fit for the cell# versus concentration at some time point and use the inverse of that Hill?

aarmey:
If Loewe requires that the min and max be shared among drugs, don't you have to do this? Individual drugs can be very different in their effects, but the control condition should be the same, and I can see it being reasonable to say that at infinite concentration the cell # is shared (same Emax).

Farnazmdi:
From what I understand I should do this:
1. we are only applying the Loewe to the cell numbers, not rates.
2. we should add a very large concentration to the end of the concentration array for each drug and set the cell number for that to 0.
3. find the best Hill fit the cell number vs. concentration curve so that we could define an inverse function.
4. apply the Loewe to these cell number curves and find the expected cell number for each pair of concentrations.
Is that right?

aarmey:
1. Seems right.
2. Don't you just need to have the minimum cell number be shared between drugs, not necessarily 0?
3. Do you need a Hill curve to do this? If your model is monotonic w.r.t. drug concentration, then you should be able to use the actual model and go from cell # to drug concentration.
4. Yes.

Farnazmdi:
Thanks.

Farnazmdi:
Sorry, I still have a gray area. Should I fit something in this process? I mean, the cell numbers are already known. I was going to just append another fake concentration and corresponding cell number to each drug. Does it mean that the fake cell number I add for all of the drugs has to be shared? I just pick the same cell number for all or there is some fitting here?

aarmey:
Hmm... I’m getting myself confused going back and forth on this.

Overall, we want to have a Loewe model of additivity, and compare the cell cycle model predictions to that additive model. So, given that, the Loewe model should **not** use the cell cycle information. The cell cycle model has reasonable handling for combining drugs at the moment.

Ok, with this overall goal, it looks like I was wrong to say you’re not fitting a Hill function. You need the Hill to describe the dose-response relationship of each drug on its own. The Loewe additivity model should then use these two Hill curves to make a prediction for additivity. The Loewe model does **nothing** with rates, only numbers of cells.

Does this help answer everything?

Farnazmdi:
Thanks, I guess everything is clear now.

Farnazmdi:
It is fixed, I guess. Now the Bliss results are being compared to the Loewe additivity to the cell numbers.
The result is in the combination.jmd
In another PR will take care of the different replicates.

Farnazmdi:
Sorry for pushing changes after tagging you. I thought it is done, then I realized I only have to work with the experimental data, not the simulations.
It should be fixed now.

Project.toml
```diff
@@ -11,10 +11,12 @@ DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
 ExponentialUtilities = "d4d017d3-3776-5f7e-afef-a10c40355c18"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+LsqFit = "2fda8390-95c7-5789-9bda-21331edee243"
```
aarmey, line 14: Did you end up using `LsqFit.jl`?

Farnazmdi: No, I used bboptim. Will remove it right away.

Farnazmdi:
I just replaced the reference model with Loewe. Here is how the fold change looks like:
image

and the combinations themselves, look like this: here each line is a specific concentration of doxorubicin with all concentrations of lapatinib. 
image

PR #298, opened 2020-06-20
average of replicates

Farnazmdi:
I added the "separate fitting to average" and "at once fitting to average" to the existing files just to have them around.

I have not yet deleted those functions and files that are related to the three replicates. Are we going to use only the average of replicates from now on, or may need the functions for separate replicates some time down the road? Otherwise, I can get rid of them.


Project.toml
```diff
@@ -10,12 +10,15 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
...
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
+IJulia = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Measures = "442fdcdd-2543-5da2-b0f3-8c86c306513e"
 Optim = "429524aa-4258-5aef-a3af-852621145aeb"
 Plotly = "58dd65bb-95f3-509e-9936-c39a10fdeae7"
 Plots = "91a5bcdd-55d7-5caf-9e0b-520d859cae80"
 Profile = "9abbd945-dff8-562f-b5e8-e1ebf5ef1b79"
 Roots = "f2b01f46-fcfa-551c-844a-d8ac1e96c665"
+StatPlots = "60ddc479-9b66-56df-82fc-76a74619b69c"
```
aarmey, line 21: Did you mean to add both StatPlots and StatsPlots?

Project.toml
```diff
@@ -10,12 +10,15 @@ Calculus = "49dc2e85-a5d0-5ad3-a950-438e2897f1b9"
 DSP = "717857b8-e6f2-59f4-9121-6e50c889abd2"
 ExponentialUtilities = "d4d017d3-3776-5f7e-afef-a10c40355c18"
 GR = "28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71"
+IJulia = "7073ff75-c697-5162-941a-fcdaad2a7d2a"
```
aarmey, line 13: Don't need Julia in the package.

Project.toml
Farnazmdi: I ended up not using it. I will remove it and IJulia in a second.

Farnazmdi: should be all set now @aarmey.

PR #302, opened 2020-06-26
Multi state

Farnazmdi:
`ODEjacMulti()`,
`Multipredict()`,
`Multicost()`
Work.
Will work on the optimization in another branch.
I wrote it so that each state has G1 and G2, separately.
** All the functions in this PR are for 2-state model.
I thought it makes sense to start with 2-state and then try to make it flexible to be for any number of states.


src/multiState.jl
```diff
@@ -0,0 +1,133 @@
...
+        G11 = Vector{eltype(p)}(undef, length(t))
+        G12 = Vector{eltype(p)}(undef, length(t))
+        G21 = Vector{eltype(p)}(undef, length(t))
+        G22 = Vector{eltype(p)}(undef, length(t))
+
+        for ii = 1:length(G11)
+            G11[ii] = sum(view(v, 1:nG1)) + sum(view(v, (2*nG1 + 2*nG2 + 1):(2*nG1 + 2*nG2 + nD1)))
+            G12[ii] = sum(view(v, (nG1 + 1):(nG1 + nG2))) + sum(view(v, (2*nG1 + 2*nG2 + nD1 + 1):(2*nG1 + 2*nG2 + nD1 + nD2)))
+            G21[ii] = sum(view(v, (nG1 + nG2):(2*nG1 + nG2))) + sum(view(v, (2*nG1 + 2*nG2 + nD1 + nD2 + 1):(2*nG1 + 2*nG2 + 2*nD1 + nD2)))
+            G22[ii] = sum(view(v, (2*nG1 + nG2 + 1):(2*nG1 + 2*nG2))) + sum(view(v, (2*nG1 + 2*nG2 + 2*nD1 + nD2 + 1):(2*nG1 + 2*nG2 + 2*nD1 + 2*nD2)))
```
Farnazmdi, line 117: This way, we need to have the time-series data for g1 in states, g1 in states, g2 in state 1 and g2 in state 2. Is this what you thought about, too? I mean, if we only have one G1 and one G2 and want the model to find states out of this data, then we only should have 1 G1 and 1 G2. But I don't know how the model is supposed to learn.

Farnazmdi:
It is failing because of this:
image

I have no idea what might be wrong.

Farnazmdi:
Here is the jacobian matrix for 2 state model with nG1 = nG2 = 2, nD1 = nD2 = 1
image

aarmey:
Ideally this would use the pre-existing Jacobian code to calculate each state's equations... will fix that error right now.

PR #328, opened 2020-10-04
edit makefile

Farnazmdi:


Farnazmdi:
@aarmey I checked this file and it should build. This error shows up in the build. I searched for it but couldn't find a solution for it. Could you please take a look at it?
image

aarmey:
Meaning you could get it to build on your account?

What are the rest of the changes here?

Farnazmdi:
Those changes are from before late August when I was editing the figure ylims, ... I don't think they mess with anything. 
I could run the code without error in the notebook. I honestly don't remember how to make `.jmd` file locally.
(I mean it builds as appose to one of the files that had errors in it and the figures were not being plotted in the pdf output)

aarmey:
Fixed! Looks like I needed to install something on the server.

makefile
```diff
@@ -1,6 +1,6 @@
 
-all: notebookPlots.pdf temporal_combination.pdf replicatesAtOnce.pdf combination.pdf avgRepsAllDrugs.pdf separateDrugsAvg.pdf
-
+all: combination.pdf
```
aarmey, line 2: Think you want to revert this?

aarmey (changes requested):
Just one edit then this is good to merge.

makefile
Farnazmdi: Thank you very much! I will revet it.

Farnazmdi:
image
It still doesn't show any artifacts. It looks like it expects svg files as output. I can make figures instead of pdfs (by looking how you have done it in a mother repo). Please let me know if I should.

aarmey:
Just fixed that.

Farnazmdi:
Sorry, they are not showing up in the artifacts yet. However, it may not be our priority now; I can make the pdf files locally by running `make all`.
If you could approve, I will merge this and try to find a strategy for outputting the optimum combination in another branch.

aarmey:
Made one more attempt at getting them to work, but yes, you can merge regardless. If it doesn't work please open an issue and tag me.

Farnazmdi:
Thank you. I need your approval to be able to merge. I think I don't have permission.

aarmey:
Just made you an admin so you should be able to override.

Farnazmdi:
Thank you!

PR #331, opened 2020-10-06
Bliss at EC50

Farnazmdi:
I know I have made lots of changes, but it was mostly bringing back what we have before that I had removed, and were needed for all the plots (we may want to get rid of them later if we are only going to work with the average of the three trials).

Could you please review the new file `Bliss.jmd` that I have added? Here is what I've done:
1. I ran the `bboptim` fitting again and restricted the bounds of progression and death rates to be [0, 1]. This took almost all day, but now we are actually satisfying the requirement of Bliss combination (which is values between [0, 1]) and as a benefit, we have finite time-series estimation for the combination of pairs of drugs, and we don't need scaling the parameters to get the combination. Long story short, the parameters have changed to a better set.

2. I extracted EC50s that have been estimated (slightly different than what I sent you yesterday). And calculated the value of each rate (i.e., each quantified drug response) at their EC50. Values at EC50 are 0.5 * (max_of_parameter + min_of_parameter). I got this based on the hill formula. Then I calculated the Bliss combination for each pair of drugs at their EC50. Since we have 5 drugs, we'll have 10 pairs for combination.

3. Then the cell number vs. time was calculated for each combination and shown in the `Bliss.jmd`.

I hope my way of calculating the parameter at EC50 and the Bliss after that is right. I appreciate it if you could confirm that.
After this, I will plot the result for Bliss applied only on pure cell number data, and then will work on Loewe results. (I think I'll do it in another branch, this branch already has 400 new lines.)


Bliss.jmd
```diff
@@ -0,0 +1,168 @@
+Left figures: Bliss applied on pure cell number of data, i.e., the traditional approach.
+Right figures: Bliss applied on the model results, i.e., our approach for drug combination.
```
aarmey, line 2: Does this note match the figures?

bliss.ipynb
```diff
@@ -0,0 +1,221 @@
+{
+ "metadata": {
```
aarmey, line 2: Is the `jmd` file supposed to replace this?

aarmey (changes requested):

Bliss.jmd
Farnazmdi: I meant to do it, but I thought I'd do it in another PR, since there are ~400 lines changed here.

bliss.ipynb
Farnazmdi: Yes, sorry, I'll delete that right away.

Bliss.jmd
Farnazmdi: Should I add it to this PR?

aarmey: No, I like starting another PR. Just wanted to check. I'll approve this.

aarmey (approved):

PR #336, opened 2020-10-10
updating combination.jmd

Farnazmdi:
`combination.jmd` shows the two cases we talked about yesterday.
Left figures show Bliss_on_parameters over Bliss_on_data. The right figures show Bliss_on_parameters over Loewe_on_data.
I restricted the maximum of the fold change shown in the heat map, now it shows the differences don't go beyond ~1.8. Generally the two models show similar patterns.
Please let me know if this is what you intended or not.

Quick note. I need to run the fitting again tonight because I had forgotten to set the upper_bound for `min_G1_prog` to 1 (it is 3 and should be 1) so the results may change a little bit.


aarmey:
I'm not sure I understand what you mean by restricting the range of differences shown. How is the model result affected by restricting the color scale?

Also, when you say the two models are mostly similar, do you mean comparing Bliss to Lowe, or cell cycle to non-cell cycle?

Farnazmdi:
I ran the model again. Now all the death and progression parameters are forced to be between 0 and 1.
Updated other files based on these parameters (those files that use the average of the three trials).
And plotted the Bliss_on_model, Bliss_on_data, and Loewe_on_data separately, with the heatmap showing cell numbers (between 0 to 30).

aarmey:
Why do the death parameters have to be between 0 and 1?

Farnazmdi:
Sorry... You are right. I will run it again it should be done by afternoon.

Farnazmdi:
The cost function reached 2010 with a wider bound for death rates (between 0 to 5), but it had reached ~1800 with the bounds being [0, 1]. Should I keep the last set of parameters or replace them with the new ones?

aarmey:
What was the cost of the fit from before you restricted the bounds?

aarmey:
Whatever has lowest cost should be used. I’m also fine using the one drug fits.

Farnazmdi:
I can replace them with parameters from one-drug-fit, but their nG1, nG2, ... are not going to be the same. Is that alright?

aarmey:
That'd be fine.

Farnazmdi:
Should I get the average of their number of phases for their combination?

Farnazmdi:
I will use those that have been estimated when we fitted all the drugs at once.

aarmey:
Oh oops. For combinations they will have to have the same phase numbers.

Farnazmdi:
There is one problem, and that is, in 3 of the combinations, the combined value for at least one of the parameters has become negative.

aarmey:
That can’t happen if Bliss combination is defined and scaled properly

Farnazmdi:
You are right. I had written some weird normalizing before applying the Bliss on progression rates. Now I removed that and don't have an assertion error anymore. If it builds I think it is ready to review.

Farnazmdi:
I think that scaling was supposed to work on those parameters that were increasing over concentrations, but they were applied to the parameters all the time. I can write some if statement to take care of it, but I'm not sure I understood it correctly.
I will write that in a new file, so tomorrow we can decide which one to show.
Please let me know if my understanding was correct whenever you got a chance.
Thank you.

Farnazmdi:
I couldn't do it because when I want to scale them back, if we are combining an increasing-parameter with a decreasing-parameter, then I can't decide if I have to scale the answer back or not, because the increasing-parameters were scaled but the decreasing ones have not.
So I discarded the changes and kept the original formula that we use for Bliss which is comb(a,b) = a+ b - a*b.

aarmey:
For Bliss, before applying it you scale such that the control condition is 1.0 and complete inhibition is 0.0. This means that, if a drug inhibits growth its scaled parameter is between 0.0 and 1.0. If a drug increases growth, the number should be greater than 1.0. After applying Bliss you then need to reverse this scaling.

Bliss should **always** be applied to the growth rates because they can be inhibited by drug relative to the control condition.

Farnazmdi:
Our scaling is not doing that, what it is doing is `parameter(control) - parameter(all_conditions)`, this way the control becomes 0.
This is my understanding of what I should do about scaling:
1. The control condition has to be 1 for either decreasing or increasing progression rates. 
2. If the progression rate is increasing over concentrations, then it should go above 1. *** Doesn't this have a conflict with the definition of Bliss which is defined for values between 0 and 1?***
3. If the prog. rate is decreasing, it should go below 1. If we have complete inhibition, it should be 0.

aarmey:
https://onlinelibrary.wiley.com/doi/epdf/10.1111/j.1744-7348.1939.tb06990.x

PR #339, opened 2020-10-13
control conditions same

Farnazmdi:


aarmey:
What is this supposed to do?

Farnazmdi:
It makes the edges in the heatmap plot the same for Loewe and Bliss.
It used to calculate the combination for `control_drugA + concentrations_drugB` like any other combination, but I updated those indices to only show cell numbers corresponding to `concentrations_drugB`. 
(I sent you the pdf of the result in the slack)

Farnazmdi:
Oops.. I need to do that for the cell cycle model, too.

aarmey:
Ok. Ping me here and send me the new document on Slack when that's done.

Farnazmdi:
In loewe, we have added one concentration t the end (10000) to force the curve to converge to zero (otherwise it cannot find a solution to the equation of loewe...). For each drug we have to provide Hill parameters [`IC50, min, max, steepness]` to calculate loewe. With our normalization, `min` becomes zero, and according to the extended array of concentrations, `max` becomes 1. I have tried a couple of other options, but the solver fails to find a root.
I think this is the reason that Loewe and Bliss are not fully similar in their edges.
@aarmey I don't have a good way to fix this, I would appreciate any comments.
In the meantime, I will plot the two conditions that are weird. I mean gemcitabine and palbo, also gemcitabine and taxol.

aarmey:
Let’s just get rid of Lowe.

Farnazmdi:
I plotted gemcitabine + paclitaxel and gemcitabine + palbo in time series and parameters, separately and in combination. The results are on page 11 of `Bliss.pdf`. I will focus on that tomorrow to see if there is a reason for the unusual behavior.

aarmey:
What is the remaining unusual behavior?

Farnazmdi:
At low concentrations, cell number is lower than the combination in high concentration.
image
I will send you the time-series and parameters in Slack.

combination.jmd
```diff
@@ -53,170 +53,123 @@ pal = DrugResponseModel.getODEparams(pal_before, concs[:, 5]);
...
 ```
+
 ## Lapatinib and doxorubicin
 
 ```julia; echo=false
 bliss_comb = DrugResponseModel.fullCombinationParam(lapat, dox, effs, 8)
-bliss_comb_cellnum = BlissModelComb(bliss_comb, g0)
-
-loeweOnData = DrugResponseModel.loweCellNum(concs, 1, 2, g1avg, g2avg);
+bliss_comb_cellnum = BlissModelComb(bliss_comb, g1avg, g2avg, 1, 2)
```
aarmey, line 67: If `BlissModelComb` is a model prediction, why does it need the data?

aarmey: If it's for scaling then pass the scale, not the entirety of the data. It's important to reduce the amount of data passed to a function to exactly what it needs, to ensure it's not dependent upon things you don't expect.

src/combination.jl
```diff
@@ -40,11 +43,37 @@ function fullCombinationParam(origP1::Array{Float64, 2}, origP2::Array{Float64,
     return fullparam
 end
 
+""" This function calculates cell number for parameter sets that are the result of Bliss on prog. rates. """
+function BlissModelComb(bliss_comb, g1avg, g2avg, id1, id2)
```
aarmey, line 47: What are `id1` and `id2`? It doesn't look like you ever use them.

src/combination.jl
```diff
@@ -98,31 +127,123 @@ end
...
     end
+
     @assert(all(combined .>= 0.0))
     return combined
 end
 
+
+####--------------- Loewe additivity ------------------####
+""" Find the inverse of a hill function (concentration), given the parameters and the effect. """
+function inv_hill(p::Array{Float64, 1}, y)
```
aarmey, line 159: I thought you removed the Loewe additivity? Moving the functions creates a huge diff.

src/combination.jl
```diff
@@ -1,5 +1,6 @@
 """ This file contains all the functions related to Bliss and temporal combinations. """
 
+######---------------- Functions for Bliss combination ----------------########
 function BlissCombination(p1::Array{Float64, 2}, p2::Array{Float64, 2}, n::Int)
```
aarmey, line 4: I'd make a function that combines `BlissCombination` and `BlissModelComb`, then test that.

src/combination.jl
```diff
@@ -15,11 +16,13 @@ function BlissCombination(p1::Array{Float64, 2}, p2::Array{Float64, 2}, n::Int)
     param2[3, :] .= p2[3, :]
     param2[4, :] .= p2[4, :]
```
aarmey, line 17: Aren't these parameters already only for specific concentrations? Does this function work if you set `n` to values other than 8?

aarmey (changes requested):

Farnazmdi:
I will address all of them, thank you for pointing these out.

Farnazmdi:
I extended the test we had, it works. And cleaned our files, so now we have two new files named "sensitivity" and "replicates" and moved the code related to them. Also removed all the code about loewe.
There is something about gemcitabine that breaks things. In any weird case, gemcitabine is involved. I ran the fitting for gemcitabine again for 1E6 (instead of 1E5) to see if the parameters weren't good, but it didn't make any difference.
I don't know where to look. I'm really sorry, it shouldn't have been such time-consuming and complicated.

src/combination.jl
```diff
@@ -98,31 +127,123 @@ end
...
-        combined[j, :, 4] = -(num[:, 1] .+ num[j, 5] .- (num[:, 1] .* num[j, 5]) .- 1.0) .* base # lap w/ palb
-        combined[j, :, 5] = -(num[:, 2] .+ num[j, 3] .- (num[:, 2] .* num[j, 3]) .- 1.0) .* base # dox w/ gem
-        combined[j, :, 6] = -(num[:, 2] .+ num[j, 4] .- (num[:, 2] .* num[j, 4]) .- 1.0) .* base # dox w/ pac
-        combined[j, :, 7] = -(num[:, 2] .+ num[j, 5] .- (num[:, 2] .* num[j, 5]) .- 1.0) .* base # dox w/ palb
-        combined[j, :, 8] = -(num[:, 3] .+ num[j, 4] .- (num[:, 3] .* num[j, 4]) .- 1.0) .* base # gem w/ pac
-        combined[j, :, 9] = -(num[:, 3] .+ num[j, 5] .- (num[:, 3] .* num[j, 5]) .- 1.0) .* base # gem w/ palb
-        combined[j, :, 10] = -(num[:, 4] .+ num[j, 5] .- (num[:, 4] .* num[j, 5]) .- 1.0) .* base # pac w/ palb
+        # the base case for either of combinations is drugA to scale the cell numbers back.
+        combined[j, :, 1] = -(num[:, 1] .+ num[j, 2] .- (num[:, 1] .* num[j, 2]) .- 1.0) .* ((g1s[T, 1, 1] + g2s[T, 1, 1]) + (g1s[T, 1, 2] + g2s[T, 1, 2])) / 2 # lap w/ dox; meaning dox changes with rows and lap changes with columns
```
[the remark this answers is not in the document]
Farnazmdi: For 5 drugs, we could have 10 pairs. The third dimension is that. The indexes in the `num` array, are the number associated with each drug (1: lapatinib, 2: doxorubicin, 3: gemcitabine, 4: taxol, 5: palbo).
I can make a function to do it only for one pair of drugs, but now I only run it once for the whole `combination.jmd` file.

src/combination.jl
Farnazmdi: I had the option before that could plot parameters for any number of points between the min and max of the concentration. But I don't know if we will need that option later or not.

src/combination.jl
Farnazmdi: Done.

src/combination.jl
Farnazmdi: Edited.

combination.jmd
Farnazmdi: Edited. Now it is only getting `bliss_comb, g0` as input.

src/combination.jl
```diff
@@ -1,37 +1,43 @@
...
     """ For 8x8 combination of drug concentrations for G1 progression rate, G2 progression rate, and death rates in G1 and G2, respectively. """
     combined = zeros(n, n, 4)
     for j = 1:n
-        for k = 1:n
-            combined[j, k, 1:2] .= (1.0 .- (param1[1:2, j] .+ param2[1:2, k] .- param1[1:2, j] .* param2[1:2, k])) .* p1[1:2, 1, 1]
+        for k = 1:n # param2 is changing, param1 is constant
+            combined[j, k, 1:2] .= (1.0 .- (param1[1:2, j] .+ param2[1:2, k] .- param1[1:2, j] .* param2[1:2, k])) .* ((p1[1:2, 1] .+ p2[1:2, 1]) ./ 2)
             combined[j, k, 3:4] .= param1[3:4, j] .+ param2[3:4, k]
         end
     end
```
Farnazmdi, line 29: I thought my indexing might be wrong like I had flipped some `j` or `k` but I wrote it down for a few columns and rows and it seemed right.

test/runtests.jl
```diff
@@ -2,7 +2,7 @@ using Test
 using Profile
 using DrugResponseModel
 
-include("testODE.jl")
-include("testHill.jl")
```
aarmey, line 6: Don't think you want to do this.

aarmey (changes requested):
One change, then please send me what the heat maps look like.

test/testCombination.jl
```diff
@@ -1,58 +1,31 @@
...
+    p2[2, 1] = 1.0          # g2 prog. rate in control
+    p2[2, 2:end] .= 0.5     # g2 prog. rate in all other concentrations
+    p2[1, 1] = 0.6          # g1 prog. rate in control
+    p2[1, 2:end] .= 0.3     # g1 prog. rate in all other concentrations
+    combination = DrugResponseModel.CombinationParam(p1, p2, 8)
+    @assert(all(combination[1:2, 2:end, 2] .>= 0.3))
+    @assert(all(combination[1:2, 2:end, 2] .<= 0.5))
 end
+
+@testset "more combination tests" begin
```
aarmey, line 20: What does this test? A more descriptive note here will be helpful.

test/testCombination.jl
```diff
@@ -1,58 +1,31 @@
 @testset "Combination tests" begin
```
aarmey, line 1: What does this test?

Farnazmdi: It tests combination for two drugs and sets their progression rates manually as you said in Slack. In drugA, g1 prog. rate in control is 1.0 and decreases to 0.8 for other concentrations; in drugB g1 prog. rate in control is 1.0 and decreases to 0.5. After we scale and calculate the Bliss and rescale back, the combined effect must be between 0.3 and 0.5 as the assert tests.
I made two more cases where the effect in end concentration is higher than control in both drugs we combine (like what we have in some cases) and when one is increasing and one is decreasing.

aarmey (approved):

Farnazmdi:
I think it is easier if I close this and after we merged your branch I can add the edits to the test and some organizing. There will be lots of conflicts if I merge this.

aarmey:
The conflicts don't look too bad.

PR #341, opened 2020-10-15
Catch errors on weave building, various cleanup

aarmey:


Farnazmdi (changes requested):
Thank you!
I caught one thing in my code today. In `combination.jl` where we calculate the Bliss on data, the x and y need to be flipped. I can do that in your branch if that's alright.

Farnazmdi (approved):

PR #342, opened 2020-10-16
more edits

Farnazmdi:
The problem was that I was using the estimated parameters from fitting drugs separately and using the nG1, nG2, nD1, and nD2 from fitting-all-at-once. I used all the parameters from the all-at-once fitting and it fixed!


aarmey (commented):
Yes! This is much better! My only remaining question is about the axis labels—don't the drugs vary in the concentration used?

aarmey (approved):
While I still have that question, this is good to merge in the mean time.

Farnazmdi:
If you mean their concentration range, it varies but they are close since it is in log. On the other hand, lapatinib, doxorubicin, and palbociclib have literally the same concentration range. If your question is about something else, could you please elaborate?

aarmey:
Oh! You're absolutely right, they are different. In that case this looks perfect. If it's alright with you I'll email this to Sean and Laura tomorrow morning with an explanation.

Farnazmdi:
That sounds perfect. Thank you so much. Please let me know if I can help with writing a draft for the explanation.

PR #350, opened 2020-10-27
5 combination time series

Farnazmdi:
Plotted %G1 versus time for 96 hours for 5 combinations from the ones Dr. Gross had chosen.


aarmey (changes requested):
The cell #—G1, G2, and total—would be much more useful.

Farnazmdi:
I plotted both of them. One the left, G1#, G2#, and total#; on the right %G1. I remember that Dr. Gross said because their data is in %G1 it is helpful, too.

aarmey (approved):
Looks good. You can email this out.

PR #358, opened 2020-11-14
A bunch of reorganization

aarmey:
This unifies separate and combined fitting into one procedure, and brings back gradient-based optimization with forward autodiff now working. We can use multi-start methods to make this more of a global search, and multi-processing if we need to speed it up.

This is likely to scale much better if we expand to multiple populations, for example.


Farnazmdi:
Ohh sorry... I must have missed this.

Farnazmdi (approved):
I think I mostly understood. Thank you.

PR #359, opened 2020-11-14
Smaller changes

aarmey:
Doing some cleanup—want to make sure this isn't breaking anything.


Farnazmdi (approved):
Thank you very much. I checked the pdf files, everything is as before.

PR #361, opened 2020-11-15
Move over some cleanup

aarmey:
I tried to simplify these two functions using broadcasting, but certainly may have broken something. Could you check? This is the last PR before I can merge some changes to the fitting process.

### DIFF
--- a/src/allDrugs.jl
+++ b/src/allDrugs.jl
@@ -186,7 +186,15 @@ end
 
 """ To find IC50 or IC90 for each drug, separately."""
 function find_IC(population, which)
-    population = abs.(population[189, :, :] .- which * population[189, 1, :])
-    IC = argmin(population, dims=2)
-    return IC[1], IC[2], IC[3], IC[4], IC[5] # returns the argument
+    lap = Array(population[189, :, 1])
+    dox = Array(population[189, :, 2])
+    gem = Array(population[189, :, 3])
+    tax = Array(population[189, :, 4])
+    pal = Array(population[189, :, 5])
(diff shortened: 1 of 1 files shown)


src/allDrugs.jl
```diff
@@ -195,15 +186,7 @@ end
...
-    pal = Array(population[189, :, 5])
-    IC_lap = argmin(abs.(which * lap[1] .- lap)) #6
-    IC_dox = argmin(abs.(which * dox[1] .- dox)) #3
-    IC_gem = argmin(abs.(which * gem[1] .- gem)) #6
-    IC_tax = argmin(abs.(which * tax[1] .- tax)) #4
-    IC_pal = argmin(abs.(which * pal[1] .- pal)) #5
-    return IC_lap, IC_dox, IC_gem, IC_tax, IC_pal # returns the argument
+    population = abs.(population[189, :, :] .- which * population[189, 1, :])
+    IC = argmin(population, dims=2)
+    return IC[1], IC[2], IC[3], IC[4], IC[5] # returns the argument
```
Farnazmdi, line 191: I see this error:
`Error: DimensionMismatch("arrays could not be broadcast to a common size; g
ot a dimension with lengths 8 and 5")`
I thought it might be from this function.

Farnazmdi (changes requested):
Thank you so much, it looks much cleaner.

src/allDrugs.jl
aarmey: Ok, I just reverted this.

PR #362, opened 2020-11-16
More cleanup

aarmey:
Ok, this is mostly an example of some more widespread cleanup that's badly needed.

You should almost never need to pass identically structured arrays like `g1_1, g1_2, g1_3`. Instead, these should be concatenated along a new dimension. In fact, the G1 and G2 data is always passed around together, so these should be combined into an array with another dimension of size 2.

The consequence of these is much easier to read code. In addition to reducing the number of variables, this will allow for broadcasting to that you don't need loops with complicated indexing. If you're concerned with making sure the data is matched up, you can always add type information to make sure the Array is 2, 3, 4... dimensional, and assertions to enforce that a dimension is a specific size.

### DIFF
--- a/src/replicates.jl
+++ b/src/replicates.jl
@@ -9,7 +9,7 @@ function predict_replicates(p1, p2, p3, g0)
 
     for i = 1:size(ps, 2) # concentration number
         for j = 1:size(ps, 3)
-            G1[:, i, j], G2[:, i, j], _ = predict(p1[:, i, j], g0, t)
+            G1[:, i, j], G2[:, i, j], _ = predict(ps[:, i, j], g0, t)
         end
     end
 
--- a/src/replicates.jl
+++ b/src/replicates.jl
@@ -1,36 +1,25 @@
 """ This file includes functions to calculate values related to replicates. """
 
 ### A function to predict G1 and G2 for the three replicates.###
-function predict_replicates(p1, p2, p3, concs1, g0)
-
+function predict_replicates(p1, p2, p3, g0)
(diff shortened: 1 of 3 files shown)


Farnazmdi (approved):
Understood. Thanks. I will do the clean ups.

aarmey:
Just to double check—this didn't affect the report results, right? It's definitely possible I made a typo...

src/replicates.jl
```diff
@@ -1,36 +1,25 @@
...
 
-    for i = 1:8 # concentration number
-        G1_1[:, i], G2_1[:, i], _ = predict(p1[:, i], g0, t)
-        G1_2[:, i], G2_2[:, i], _ = predict(p2[:, i], g0, t)
-        G1_3[:, i], G2_3[:, i], _ = predict(p3[:, i], g0, t)
+    for i = 1:size(ps, 2) # concentration number
+        for j = 1:size(ps, 3)
+            G1[:, i, j], G2[:, i, j], _ = predict(p1[:, i, j], g0, t)
+        end
     end
```
Farnazmdi, line 15: Oh sorry, there is one. `Error: BoundsError: attempt to access 9×@*(8 Array(*@{Float64,2} at index [1:9, 1
, 2]` there is an error in the `replicatesAtOnce.jmd` which I think may refer to this part. Maybe the `effs1,2,3` that is passed to the `predict_replicate` is not 3dimensions?

Farnazmdi:
I checked that file, it has been fixed.

PR #364, opened 2020-11-18
Bliss clean

aarmey:
I updated the build process so that it fails if there are problems within the Weave notebooks. I think the only remaining problem is `notebookPlots.jmd` has an error. Could you take a look at this? I also deleted one notebook because every function it called wasn't present in the package.


Farnazmdi:
Sure, will do right away.

Farnazmdi:
Is it alright if I make changes to this branch?

aarmey:
Yep.

Farnazmdi:
The problem I guess is that nG1, nG2, nD1, nD2 sometimes might be less than 10 and in this case, the range we defined to vary the parameter for sensitivity analysis (from 0.1 to 10 times the value) becomes less than 1 and creates 0 or 1 and when this propagates to the ODEjac and wants to create an array with size 0 or -1. I modified the range we vary for these 4 parameters and I think it should fix this error.

aarmey:
@Farnazmdi why not, rather than making a special case in the sensitivity function, just move up the step parameters used in the sensitivity notebook to avoid this?

Farnazmdi:
Sure, sounds better. Will do.

Farnazmdi (approved):
I think the notebookPlots.jmd is fixed now. Thank you very much for the clean ups.

PR #369, opened 2020-11-19
Some cleanup

aarmey:
I think this all works fine—just thought it'd be helpful to show you the trick for applying the same style across multiple plots. This should help clean up some of the plots a bit more.


Farnazmdi (approved):
Got it, thank you.

PR #370, opened 2020-11-20
Sensitivity

Farnazmdi:


src/sensitivity.jl
```diff
@@ -54,3 +54,28 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
 end
+
+""" This function calculates the simulated cell number for a pair of drugs in their exact concentration.  """
+function calc_cellNum(pDr1, pDr2, g0)
+    Dr1CellNum = predict(pDr1, g0, 96.0)
+    Dr2CellNum = predict(pDr2, g0, 96.0)
+    normNum = 1.0 .- [Dr1CellNum/g0, Dr2CellNum/g0]
+    combin = -((normNum[1] + normNum[2] - normNum[1] * normNum[2]) .- 1.0) * g0 
+    return combin
+end
```
Farnazmdi, line 65: I wrote these functions based on only one pair of concentrations so that the answer (the difference) will be a scalar number.

src/sensitivity.jl
```diff
@@ -54,3 +54,28 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+
+fd(x) = calc_diff(x, Dr2_params, effs, combin, concs, conc1_indx, conc2_indx, g0)
+
+function get_derivative(x, combin, concs, Dr1_indx, Dr2_indx, k)
+    # "k" is the index of the parameter we want to calculate the derivative of diff with respect to it.
+    fd(x) = calc_diff(x, combin, concs, Dr1_indx, Dr2_indx) # closure?
+    out = DiffResults.DiffRresult(x[k])
+    return ForwardDiff.derivative(out, fd, x[k])
+end
```
Farnazmdi, line 81: I have difficulty handling the differentiation. I couldn't find an example online that could help with the correct syntax.
I know that I should use closure to set the input of the `calc_diff` function only the parameters, that's why I defined `fd(x)` but I think clearly it is wrong. 
All the examples I saw were written without any parameter and they had an input vector where they were differentiating the function w.r.t. the input vector. I don't know how to tell the function to take the derivative of this function only with respect to one of the elements in the parameter vector. Maybe it is not necessary...

src/sensitivity.jl
```diff
@@ -54,3 +54,26 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+    bliss_comb_cellnum = BlissModelComb(bliss_comb, g0)[conc1_indx, conc2_indx]
+    return combin - bliss_comb_cellnum
+end
+
+function get_derivative(x, Dr2_params, effs, combin, concs, conc1_indx, conc2_indx, g0)
+    # "k" is the index of the parameter we want to calculate the derivative of diff with respect to it.
+    fd(x) = calc_diff(x, combin, concs, Dr1_indx, Dr2_indx, g0) # closure?
+    out = DiffResults.DiffResult(x)
+    return ForwardDiff.derivative.(out, fd, x)
+end
```
Farnazmdi, line 79: @aarmey I couldn't get around the error regarding how to call the `ForwardDiff.derivative!(...)`. I couldn't find any relevant examples online. Could you please help me with this function?

src/sensitivity.jl
```diff
@@ -54,3 +54,26 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+    bliss_comb = DrugResponseModel.fullCombinationParam(Dr1_params, Dr2_params, effs, 8, conc1_indx, conc2_indx)
+    bliss_comb_cellnum = BlissModelComb(bliss_comb, g0)[conc1_indx, conc2_indx]
+    return combin - bliss_comb_cellnum
+end
+
+function get_derivative(x, Dr2_params, effs, combin, concs, conc1_indx, conc2_indx, g0)
+    # "k" is the index of the parameter we want to calculate the derivative of diff with respect to it.
+    fd(x) = calc_diff(x, combin, concs, Dr1_indx, Dr2_indx, g0) # closure?
+    out = DiffResults.DiffResult(x)
+    return ForwardDiff.derivative.(out, fd, x)
```
aarmey, line 78:
```suggestion
    return ForwardDiff.gradient(fd, x)
```

src/sensitivity.jl
```diff
@@ -54,3 +54,26 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
+function calc_diff(Dr1_params, Dr2_params, effs, combin, concs, conc1_indx, conc2_indx, g0)
+    bliss_comb = DrugResponseModel.fullCombinationParam(Dr1_params, Dr2_params, effs, 8, conc1_indx, conc2_indx)
+    bliss_comb_cellnum = BlissModelComb(bliss_comb, g0)[conc1_indx, conc2_indx]
+    return combin - bliss_comb_cellnum
+end
+
+function get_derivative(x, Dr2_params, effs, combin, concs, conc1_indx, conc2_indx, g0)
+    # "k" is the index of the parameter we want to calculate the derivative of diff with respect to it.
+    fd(x) = calc_diff(x, combin, concs, Dr1_indx, Dr2_indx, g0) # closure?
+    out = DiffResults.DiffResult(x)
```
aarmey, line 77:
```suggestion

```

PR #372, opened 2020-11-21
jmd

Farnazmdi:
I will send you the pdf in the Slack right now.


Sensitivity.jmd
```diff
@@ -0,0 +1,75 @@
...
+    effs = getODEparamsAll(p, concs);
+    # Check that these at least run
+
+xaxis = ["lpt+dox", "lpt+gem", "lpt+tax", "lpt+plb", "dox+gem", "dox+tax", "dox+plb", "gem+tax", "gem+plb", "tax+plb"]
+function calc_ders(effs, ic50s, g0)
+    deriv = zeros(9,10)
+    q = 1
+    for i=1:4
+        for j=(i+1):5
+            deriv[:, q] = DrugResponseModel.get_derivative(effs[:, ic50s[i], i], effs[:, ic50s[j], j], effs, i, j, ic50s[i], ic50s[j], g0)
```
aarmey, line 48: These derivatives need to be w.r.t. `p`, not `effs`. `getODEparamsAll` being outside of `get_derivative` indicates this is not right...

Farnazmdi:
There is one error that if fixed, the notebook will be built. 
The input array (p_of Hill) should be a Float array for `getODEparamsAll()` but it is a Dual number. This is the problem when we pass it to the derivative function. 
Is there an easy way of fixing it?

aarmey:
Just remove the type information on the function. No need to include it.

Farnazmdi:
I can't find any place in the `getODEparamsAll` that is converting to Float64, but this error refers to line 11 of the function...
image

aarmey:
`effects` needs to be the same type as `p`.

https://docs.julialang.org/en/v1/base/arrays/#Base.zeros

Farnazmdi:
I'm really sorry, I couldn't fix it :((( 
When we are not defining a variables as `T[zeros...]` and it is just the output of some other function, like what we have in both `calc_cellNum` and `calc_diff` when we calculate the `effects` from `Hillp`, how do we make sure their type match? @aarmey

src/sensitivity.jl
```diff
@@ -69,9 +69,9 @@ function calc_cellNum(Hillp, dr1_ind, dr2_ind, g0, concs)
 end
 
 """ Calculates the difference between the bliss_cell number and bliss_params. """
-function calc_diff(Hillp, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, concs, g0)
-    effs = getODEparamsAll(Hillp, concs)
-    combin = calc_cellNum(Hillp, Dr1Ind, Dr2Ind, g0, concs)
+function calc_diff(Hillp::Vector{T}, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, concs, g0) where {T}
+    effs = T[getODEparamsAll(Hillp, concs)]
+    combin = T[calc_cellNum(Hillp, Dr1Ind, Dr2Ind, g0, concs)]
     bliss_comb = DrugResponseModel.fullCombinationParam(effs[:, :, Dr1Ind], effs[:, :, Dr2Ind], effs, 8)
```
Farnazmdi, line 75: This is the last thing I tried, but doesn't work.

aarmey:
What is the error that you get?

aarmey:
Removed some of the code changes you made. The only problem I see is you're not passing the right arguments to `get_derivative`—it's not a typing problem.

src/sensitivity.jl
```diff
@@ -56,25 +56,27 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
 end
 
 """ Calculates the difference between the bliss_cell number and bliss_params. """
-function calc_diff(Dr1_params, Dr2_params, effs, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, g0)
-    # Dr1_params/Dr2_params is a 9 x 1 Array, including the 9 parameters of ODE model. 
-    combin = calc_cellNum(Dr1_params, Dr2_params, g0)
+function calc_diff(Hillp, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, concs, g0)
+    effs = getODEparamsAll(Hillp, concs)
+    combin = calc_cellNum(effs, Dr1Ind, Dr2Ind, g0)
     bliss_comb = DrugResponseModel.fullCombinationParam(effs[:, :, Dr1Ind], effs[:, :, Dr2Ind], effs, 8)
```
aarmey, line 74: Shouldn't `Dr1Ind` and `conc1_indx` always match?

src/sensitivity.jl
```diff
@@ -56,25 +56,27 @@ function plotUnitSensitivity(paramRange, result, realParam, i)
...
     return combin
 end
 
 """ Calculates the difference between the bliss_cell number and bliss_params. """
-function calc_diff(Dr1_params, Dr2_params, effs, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, g0)
-    # Dr1_params/Dr2_params is a 9 x 1 Array, including the 9 parameters of ODE model. 
-    combin = calc_cellNum(Dr1_params, Dr2_params, g0)
+function calc_diff(Hillp, Dr1Ind, Dr2Ind, conc1_indx, conc2_indx, concs, g0)
+    effs = getODEparamsAll(Hillp, concs)
+    combin = calc_cellNum(effs, Dr1Ind, Dr2Ind, g0)
```
aarmey, line 73: Why are you passing the indices in just to subset `effs`? This violates the principal of providing a function with only the data it needs.

src/sensitivity.jl
Farnazmdi: Dr1Ind is between 1 to 5 and it refers to the name of the drug (1=lapatinib, 2=dox,...) the conc1_indx is the index of the concentration for that drug.

src/sensitivity.jl
Farnazmdi: Fixed with the last commit.

Farnazmdi:
The error is 
image
I tried to change some things in the `predict()` but they didn't work. It was the start of the errors.

aarmey:
Try now? There's a special case for just one timepoint in predict, and it wasn't setup to handle Duals.

Farnazmdi:
Thank you. I made the type of other variables that needed to be the same, but it keeps erroring the `combined` 3D array that is defined inside the `CombinationParam()`. I made the type of it the same as the input parameter, too, but it still errors and it is not clear what's wrong.
image

aarmey:
Your array constructors are not the right syntax. Their type has both the element type and dimensions. You can use the Matrix constructor if you only want to provide the type. See here:

 https://docs.julialang.org/en/v1/base/arrays/

Farnazmdi:
Thank you very much. Changed it to Array and looks working.
I will send you the pdf file in Slack in a minute.

Sensitivity.jmd
```diff
@@ -0,0 +1,135 @@
...
+```julia
+concs, populations1, g1s1, g2s1 = load(189, 1);
+concs, populations2, g1s2, g2s2 = load(189, 2);
+concs, populations3, g1s3, g2s3 = load(189, 3);
+g0 = g1s1[1,1,1]+g2s1[1,1,1];
+
+g1avg = (g1s1 .+ g1s2 .+ g1s3) ./ 3; # pure data
+g2avg = (g2s1 .+ g2s2 .+ g2s3) ./ 3; # pure data
+popavg = g1avg .+ g2avg;
+IC50_lap, IC50_dox, IC50_gem, IC50_tax, IC50_pal = DrugResponseModel.find_IC(popavg, 0.5);
```
aarmey, line 27: This is a poor way to find the IC50 (it will essentially always be between the concentrations tested). Don't you have a better IC50 estimate from the parameter fits?

Sensitivity.jmd
```diff
@@ -0,0 +1,135 @@
...
+ic50s = [IC50_lap, IC50_dox, IC50_gem, IC50_tax, IC50_pal]
+```
+
+```julia
+p = [97.7434, 1.41726, 0.0501761, 0.930205, 0.111494, 0.123604, 0.552216,
```
aarmey, line 32: Should these parameters just be in the package, rather than repeated in every notebook? This can be done on a separate PR.

Sensitivity.jmd
```diff
@@ -0,0 +1,135 @@
...
+        92.9045, 1.40024, 0.996396, 0.0513087, 0.521831, 0.6535, 0.566578, 
+        15.5317, 2.3689, 0.592433, 0.999986, 0.0283363, 0.286975, 0.503328, 
+        3.96929, 4.62768, 0.0512281, 0.307528, 0.549714, 0.378717, 0.50959, 
+        63.4248, 0.976052, 0.16582, 0.740009, 0.0572609, 0.0776912, 0.534201, 
+        0.734513, 0.375555, 16.8387, 12.3945, 30.176, 14.5352]
+    g0 = 20.0
+
+function decide_index(Dr1Ind)
+    if Dr1Ind == 1
+        indexes1 = 1:6
```
aarmey, line 42: If `p` is a 2D array you don't need this complicated indexing.

Sensitivity.jmd
```diff
@@ -0,0 +1,135 @@
...
+
+```julia
+concs, populations1, g1s1, g2s1 = load(189, 1);
+concs, populations2, g1s2, g2s2 = load(189, 2);
+concs, populations3, g1s3, g2s3 = load(189, 3);
+g0 = g1s1[1,1,1]+g2s1[1,1,1];
+
+g1avg = (g1s1 .+ g1s2 .+ g1s3) ./ 3; # pure data
+g2avg = (g2s1 .+ g2s2 .+ g2s3) ./ 3; # pure data
+popavg = g1avg .+ g2avg;
```
aarmey, line 26: If you use the parameters for the IC50's, you never need this imported data.

Sensitivity.jmd
```diff
@@ -0,0 +1,135 @@
...
+  error: false
+  echo: false
+---
+
+```julia
+import Pkg; Pkg.instantiate()
+
+Pkg.activate(".")
+using DrugResponseModel
+using ExponentialUtilities
```
aarmey, line 12: This is unused.

aarmey (changes requested):

Sensitivity.jmd
Farnazmdi: Yes, they were different in two cases I guess. Should I use the IC50s that the model picked?

Sensitivity.jmd
Farnazmdi: Will do.

Sensitivity.jmd
Farnazmdi: I'll make a function to create the 2D from the `p` in the new branch.

Sensitivity.jmd
Farnazmdi: I'm doing this, it takes a bit longer because I need to extrapolate the Hill effects based on the EC50.

Farnazmdi: Oh.. I realized that I don't have to extrapolate... it is EC50! (please ignore me)

src/combination.jl
```diff
@@ -1,14 +1,35 @@
...
+    p2[3:4] .= pp2[3:4]
+
+    c = Array{eltype(pp1), 1}(undef, 9)
+    c[1:2] .= (1.0 .- (p1[1:2] .+ p2[1:2] .- p1[1:2] .* p2[1:2])) .* ((control[1:2, 1] .+ control[1:2, 2]) ./ 2)
+    c[3:4] .= p1[3:4] .+ p2[3:4]
+ 
+    c[5:9] .= pp1[5:9]
+    return c
+end
+
```
Farnazmdi, line 23: I will use this and edit the function that calculates the Bliss combination for parameters. I needed this because I wanted to only pass the EC50 parameters to it.

aarmey (approved):
Great.

PR #375, opened 2020-11-25
Clean

Farnazmdi:
1. Used the unit function I wrote in a previous branch for bliss on parameters and cleaned up a little bit based on that.
I also realized I had two repeated function that I deleted them.


test/runtests.jl
```diff
@@ -2,8 +2,8 @@ using Test
 using Profile
 using DrugResponseModel
 
-include("testODE.jl")
-include("testHill.jl")
-include("testDrugAll.jl")
```
aarmey, line 7: You're commenting out all the tests.

aarmey (changes requested):
Fix the testing file, and then check that the report output doesn't change in any surprising ways. Then, this should be ready to merge.

test/runtests.jl
Farnazmdi: Oh sorry.. forgot to bring them back

PR #378, opened 2020-12-03
some changes

Farnazmdi:
1. unified the ylims
2. scaled the nD1 and nD2 based on our discussion.


Farnazmdi:
@aarmey I have another branch here that compares the combination with the single drug effects. Their pattern is identical, just the magnitude is slightly different. I guess we don't want to apply that change then.

aarmey (approved):

PR #380, opened 2020-12-05
Interactive

Farnazmdi:
I checked the code in my local, it runs. 
In each cell we can manipulate the 4 Hill parameters `[max_g1_prog, max_g2_prog, max_g1_death, max_g2_death]`.
I will play with it a little bit tomorrow and will update you if I found something promising. We could look at that together, too.


src/DrugResponseModel.jl
```diff
@@ -16,6 +16,7 @@ import ExponentialUtilities
 import LineSearches
 using Polynomials
 using DiffResults
+using Interact
```
aarmey, line 19: If Interact isn't used within the package then it doesn't need to be imported or placed in the package manifest.

aarmey (changes requested):
One change to make sure this builds.

src/DrugResponseModel.jl
Farnazmdi: Should I remove changes made to the manifest.toml, too?

aarmey: What is the purpose of the manifest?

Farnazmdi: It has information about dependencies. So I'll remove the changes.

PR #385, opened 2020-12-20
fix

Farnazmdi:
predict returns negative cell numbers sometimes when we only want to calculate cell numbers for one time point. In other cases that we have a range of time, it works just fine. I edited the functions that calculate cell numbers for combinations to call `predict` with a range of time, but keep the very last data point.


aarmey:
Aha! That makes sense as the one-timepoint solving is performed differently. I'll make an alternative fix on a different branch.

PR #386, opened 2020-12-20
Fix zeros?

aarmey:
@Farnazmdi try this out and let me know if it fixes the issue.


Farnazmdi (approved):
I tried it in my local and it worked! Thank you so much.

PR #394, opened 2021-01-23
multi start in a notebook for single drugs

Farnazmdi:
@aarmey I read that it only works with NLopt for now and in the source code, I couldn't find an argument for the gradient to pass to it. 
I tried it in a notebook for only 5 data points and it didn't fail, although for the whole data it is slow. I also couldn't find any options like `verbose` so that I could see each step. 

Could you please help me with including the gradient in it?


aarmey:
Did you try the gradient based on the layout here?

https://github.com/JuliaOpt/NLopt.jl

Farnazmdi:
I will check it out right away. Thank you.

src/Hill.jl
```diff
@@ -42,19 +42,34 @@ end
...
+    end
+    P = MinimizationProblem(f, low, high)
+    opt = Opt(:LD_MMA, 20)
+    opt.min_objective = my_func
+    opt.xtol_rel = 1e-4
+    # local_method = NLoptLocalMethod(NLopt.LD_MMA)
+    multistart_method = TikTak(100)
+    p = multistart_minimization(multistart_method, opt, P)
+
+    return p.location, p.value
```
Farnazmdi, line 57: Is this close to that interface?
It errors that we cannot pass `opt` to the `mulitstart_minimization`. It should be the type `!Matched::MultistartOptimization.NLoptLocalMethod`. 
image

aarmey:
How about just copy out the multistory function and replace the local call part with your local optimization method?

https://github.com/tpapp/MultistartOptimization.jl/blob/20523fb696fcc649f5a25b64c77b81185362f2d7/src/MultistartOptimization.jl#L180

aarmey:
Then you should even be able to use Optim.jl like you had before.

Farnazmdi:
There is one problem.
In this function, the `TikTak` that I guess is the important part of multi_start, provides an initial point for optimizer; but our optimizer takes in lower and upper bounds. 
I don't know how to fix that. Should I change our optimizer so it takes initial values instead of bounds?

1.

aarmey:
Our optimizer should accept the lower and upper bounds, along with the initial point. Right now it makes up a starting point from the bounds. It's just a matter of replacing this line:

https://github.com/meyer-lab/DrugResponseModel.jl/blob/d3acb3a201436a194269e6c750c14afb75fbd911/src/Hill.jl#L43

Farnazmdi:
Ohhh... my bad. I just saw that there is initial x in the optimizer.

src/Hill.jl
```diff
@@ -43,33 +43,53 @@ function grad_helper!(G, f, range, params)
...
-""" Hill optimization function. """
-function optimize_hill(conc::Vector, g1::Matrix, g2::Matrix; maxstep = 100000)
-    f(x) = residHill(x, conc, g1, g2)
-    g!(G, x) = grad_helper!(G, f, 10:13, x)
+function sobol_starting_points(minimization_problem::MinimizationProblem, N::Integer)
+    @unpack objective, lower_bounds, upper_bounds = minimization_problem
+    s = SobolSeq(lower_bounds, upper_bounds)
+    skip(s, N)                  # better uniformity
+    map(fetch, map(x -> @spawn(LocationValue(x, objective(x))), Iterators.take(s, N)))
+end
```
Farnazmdi, line 72: @aarmey is it possible to write this line without using `LocationValue`?
 I deleted this struct (the main reason was I wasn't sure what exactly it is doing). Now, I have this error `no method matching length(::LocationValue{Array{Float64,1},Float64})` which goes back to this.

aarmey: Not sure what this is doing either. Maybe you just need to put `MultistartOptimization.LocationValue` because it's not exported?

aarmey: Do you have to copy this function out of the package too?

Farnazmdi: I tried just calling it by doing `MultistartOptimization.sobol_starting_points`, but the error occurred. I thought if I copy this function and change that specific line, the error would fix.

src/Hill.jl
```diff
@@ -43,33 +43,53 @@ function grad_helper!(G, f, range, params)
...
 
-""" Hill optimization function. """
-function optimize_hill(conc::Vector, g1::Matrix, g2::Matrix; maxstep = 100000)
-    f(x) = residHill(x, conc, g1, g2)
-    g!(G, x) = grad_helper!(G, f, 10:13, x)
+function sobol_starting_points(minimization_problem::MinimizationProblem, N::Integer)
+    @unpack objective, lower_bounds, upper_bounds = minimization_problem
+    s = SobolSeq(lower_bounds, upper_bounds)
+    skip(s, N)                  # better uniformity
+    map(fetch, map(x -> @spawn(LocationValue(x, objective(x))), Iterators.take(s, N)))
```
aarmey, line 71:
```suggestion
    map(x -> LocationValue(x, objective(x)), Iterators.take(s, N))
```

Try this.

PR #395, opened 2021-01-23
Some adjustments to make fitting faster.

aarmey:
This uses multithreading in part so you'll want to start Julia with the argument `--threads 18`. This takes ~0.7 sec per iteration.


Farnazmdi (approved):

Farnazmdi (approved):
Thank you.

PR #396, opened 2021-01-24
Simplified multi-start

aarmey:
Multi-start is just using Sobol numbers to generate initial values. This seems to work correctly here. In practice you'll want to run the local search over more iterations, and probably increase the number of initial points explored.


src/Hill.jl
```diff
@@ -45,20 +45,37 @@ end
...
 
     return Optim.minimum(results), Optim.minimizer(results)
 end
 
 
+function optimize_helper(f, g!, low::Vector, high::Vector, maxstep::Int)
+    sobolPoints = 10
+    s = SobolSeq(low, high)
+    skip(s, sobolPoints)
+
```
Farnazmdi, line 64: Thank you so much. I will continue based on this.
One small question. These two functions have the same name. I think the second is calling the first one. I can change the name if it is okay.

Farnazmdi (approved):
Thank you so much.

src/Hill.jl
aarmey: Yep. Which one gets called depends on whether you provide an `x0` argument.

aarmey: Changing the name is fine.

Farnazmdi:
Could I merge this into my branch?

PR #397, opened 2021-01-25
Edits

aarmey:
Loosening the bounds during the optimization seems to help quite a bit.


aarmey:
Not planning to merge this. Just copy whatever edits would be helpful.

Farnazmdi:
I will do the same, thanks. Doesn't this increase the run time though? When I tried doing this on the global optimization we had, it was incredibly slow.

aarmey:
Nope.

PR #401, opened 2021-01-28
Multi thread

Farnazmdi:
I followed their example where they specified a set of methods that work with multi-threading (here I used `dxnes`), and add another argument, `NThreads=Threads.nthreads()` in the BBOptim function.
However, when I ran it in the terminal (while running Julia with 18 threads) the speed was the same (~12 evaluations per second).
Could you please let me know if you noticed something that I'm missing here? @aarmey


aarmey:
Hmm... I see the same results... all I can figure is that matrix operations that call into BLAS can only be called by one thread at a time. Not sure of a way to get around this...

Farnazmdi:
Thank you. I won't merge it then.
Should I proceed with your suggestion about dividing each phase into two phases or should I fit other drug pairs and try combinations for those pairs?

aarmey:
I wonder about loosening the bounds and whether that helps with the fit. Otherwise, yes, let’s split the phases.

Farnazmdi:
Thanks. Will do. I tried that with the gradient based and let it run over night, but in the end, the fitness was ~32000 which is 4X bigger than the usual fitness cost. 
I will run it with the global optimization and update you tomorrow.

PR #406, opened 2021-02-04
Setup using StaticArrays

aarmey:
Solving won't get faster than this. ~1500 iterations / sec for all the drugs.

This also sets up the start of a starting point solver based on the control parameters. This is because, when the subphases aren't identical, cells will be distributed among them unevenly.


Farnazmdi (approved):
Thank you very much

PR #407, opened 2021-02-04
removed num phases and fixed optimzizer for drug pairs

Farnazmdi:


src/ODEmodel.jl
```diff
@@ -2,22 +2,23 @@
...
     A = diagm(0 => v1, -1 => v2)
     A[1, nSp] = 2 * p[2]
+    A[nSp, nSp-1] = p[4]
 
     A = SMatrix{nSp, nSp}(A)
     A = exp(A * t)
+    print(A, "there it is")
 
     return A
 end
```
Farnazmdi, line 24: These are my changes. 
I added a notebook called "test.ipynb" In that, I called the ODEjac(). @aarmey

Farnazmdi:
@aarmey I should have asked this yesterday, sorry, is the `startV()` function supposed to give us the initial cells, or is it an initial guess for parameters? It returns an array with the size of parameters, but the way it is used in the `predict()` is like it is an initial guess for cell numbers. And I couldn't figure out where to call it.

aarmey:
Eventually, yes, the starting vector needs to come from the control condition, and `startV` handles this for you. For now you can ignore it and get everything else working?

Farnazmdi:
Thanks.

src/ODEmodel.jl
```diff
@@ -3,18 +3,19 @@
...
+function ODEjac(p::AbstractVector{T}, t::Float64) where {T <: Real}
+    # p = [alpha1, alpha2, beta1, beta2, gamma11, gamma12, gamma21, gamma22]
+    v1 = [-ones(Int(nG1/2)) * (p[5] + p[1]); -ones(Int(nG1/2)) * (p[2] + p[6]); -ones(Int(nG2/2)) * (p[3] + p[7]); -ones(Int(nG2/2)) * (p[4] + p[8])]
+    v2 = [ones(Int(nG1/2)) * p[1]; ones(Int(nG1/2)) * p[2]; ones(Int(nG2/2 - 1)) * p[3]; ones(Int(nG2/2 - 1)) * p[4]]
 
     A = diagm(0 => v1, -1 => v2)
     A[1, nSp] = 2 * p[2]
+    A[nSp, nSp-1] = p[4]
 
     lmul!(t, A)
```
aarmey, line 20: Found some documentation that StaticArrays should only be used with arrays of <100 elements, as the stress on the compiler scales with the square of the size. This replacement should work for us.

Farnazmdi: Thank you so much.

Farnazmdi:
@aarmey I tried fitting single drugs and pairs of drugs, the fit doesn't seem good.
Single drug:
image

when pairs with another drug:
image

When paired with some drugs, the fit gets a little bit better.
But some of them are really out! Like this:

image

I tried with different numbers of nG1 and nG2, and also widened the bounds. Neither helped with the fitting.
I guess I will bring back the death phases and see if gets better. Please let me know if you have another suggestion. Thanks.

aarmey:
First, we should use the right starting vector. Second, we can further split the phases into four rates.

aarmey:
Also maybe we can get this to a point where it builds first? It’d be good to review and ensure there aren’t any bugs.

Farnazmdi:
Sure. Will do it right now.
I will edit everything based on sharing the control conditions between the two drugs.

Farnazmdi:
I fixed the tests, but for builds to pass, I need to provide the estimated parameters that work with the new structure. It will take a long time and since it didn't work well, it will be all for nothing.
I will divide each phase into 4 (instead of two) and see if it imrpoves.

aarmey:
Using the control to find the starting vector should be the first change. This absolutely has to be done, and could change your conclusion here about two parameters working.

src/ODEmodel.jl
```diff
@@ -40,12 +40,36 @@ function vTOg(v::AbstractVector)
...
+    return DrugResponseModel.optimize_helper(f, smin, smax, 100000)
+end
+
+function newPredict(p, pControl, t::Union{Real, LinRange}, g1data = nothing, g2data = nothing)
+
+    vStart = startV(pControl)
+    # Note that vStart is always scaled so the starting cell number is 1.0
+
+    return predict(p, vStart, t, g1data, g2data)
+end
```
Farnazmdi, line 63: Is this a correct way of handling it? I am getting very high cost values.

aarmey: Looks right. Is the data scaled in such a way that the cell number at t=0 is 1.0? If not this vector needs to be scaled.

Farnazmdi: Oh.. mo, this is my mistake. I will fix it, thank you.

Farnazmdi:
In this PR:
1. Bliss.jmd was removed, because it was time series simulations of combination that we have a more complete version in another file.
2. g0 was adjusted in all files based on the new changes.
3. I defined a dummy random vector of parameters for jmd files to just build, I will replace them with correct estimations after fitting and other model adjustments.
4. I defined a unified `getODEparams` that could be used everywhere and removed `getODEparamsAll` and `getODEparamsPairs`.

I manually checked the ODEjac with a set of parameters, and it was correct.
Some other simplifications were made.

We have a separate folder for functions for fitting pairs of drugs.

src/ODEmodel.jl
```diff
@@ -44,13 +48,35 @@ function vTOg(v::AbstractVector)
...
 end
 
+""" To find the initial vector. """
+function residue(p, g1, g2)
+    t = LinRange(0.0, 0.5 * size(g1, 1), size(g1, 1))
+    g00 = g1[1, 1] + g2[1, 1]
+    return predict(p, g00, t, g1[:, 1], g2[:, 1])[1]
+end
+
+function optimise_control(g1, g2)
```
aarmey, line 58: Are you fitting the control separate from everything else? Shouldn't this be fit with all the other drug conditions?

src/Hill.jl
```diff
@@ -36,32 +36,38 @@ end
...
 
     return optimize_helper(f, low, high, maxstep)
 end
 
-""" A function to convert the estimated hill parameters back to ODE parameters. """
-function getODEparams(p::Vector, concentrations::Vector{Float64})
-    effects = Matrix{eltype(p)}(undef, 9, length(concentrations))
-
+function getODEparams(p, conc, nMax)
+    effects = zeros(eltype(p), 9, length(conc[:, 1]), nMax)
```
aarmey, line 51: Can't nMax be derived from the length of `p`? If so, this would help prevent errors by not duplicating information.

aarmey (approved):
Two questions then this is ready to merge.

src/Hill.jl
Farnazmdi: fixed.

src/ODEmodel.jl
Farnazmdi: Yes, I didn't include Hill at all, and this was just to find the ODE params of the control for the purpose of finding the initial cell vector. I thought we need that first before fitting other concentrations. If this is not what you intended I can change it.

aarmey: The control should be fit at the same time as everything else. I think you can assume the death rates are 0 for the control.

Farnazmdi: Thanks, I will do this tomorrow.

Farnazmdi: Following up this comment:
We have the death rates =0 for control conditions, and we fit the control along with other concentrations. This step is just to find the control that will then be used to find the initial cells in sub-phases. I mean this will form the pControl argument for `newPredict()`:

image

Please let me know if I am doing it wrong. Thank you.

aarmey: Why does optimize control exist then?

aarmey: As far as I can tell you’re still fitting the control separately. You can’t do this. The best fit for the control will also be influenced by the drug treatment data, due to the Hill relationship.

Farnazmdi: Sorry I couldn't figure out how to do it. 
Could you please clarify a little bit about how to get the control parameters for the g0 vector?
It seems to me that we have to have the control parameters to use it as the g0 vector in `newPredict`. If we run all the conditions, we have to go it with a point initial value, not a vector.

aarmey: p are parameters ultimately calculated from the unknown parameters being fit. pControl is no different. It should be derived from the unknown vector being used in the fitting process for both drugs.

This is exactly how we had it before—the control condition was fit along with the drug effects.

aarmey: The control parameters are the parameters of the model in the absence of drug.

aarmey: I don’t know what you mean about needing an initial point. pControl is passed to startV to calculate the initial point. There’s no contradiction here.

Farnazmdi: Thank you. It should be fixed.

aarmey (approved):

PR #409, opened 2021-02-07
Example of using control-defined starting point

aarmey:
Here's an example of how to use the control condition to find the starting vector. This function takes the parameters for the control condition and the one you want to solve for. The control parameters are passed to `startV`, which gives back the starting vector.


aarmey:
@Farnazmdi FYI.

Farnazmdi:
Thank you.
I wrote a function that estimates only the 9 ODE parameters for control. Is it intended as the pControl here?

aarmey:
Yes.

PR #412, opened 2021-02-10
Simplify core model and add some additional assertions

aarmey:
@Farnazmdi I simplified the core ODE model a bit, and added an extra assertion for the parameter length. It seems to be catching a few issues. Could you take a look? Once this is passing it can be merged.


src/ODEmodel.jl
```diff
@@ -52,23 +52,15 @@ function vTOg(v::AbstractVector)
...
-
-
 """ Predicts the model given a set of parametrs. """
-function predict(p, g_0, t::Union{Real, LinRange}, g1data = nothing, g2data = nothing)
-    if g_0 isa Real
-        v = Vector{eltype(p)}(undef, nSp)
-        v[1:nG1] .= p[9] * g_0 / nG1
-        v[(nG1+1):end] .= (1.0 - p[9]) * g_0 / nG2
+function predict(p::AbstractVector, g_0::AbstractVector, t::Union{Real, LinRange}, g1data = nothing, g2data = nothing)
+    @assert length(p) == 8
```
aarmey, line 57: I got rid of `newPredict` because this can just be accomplished in `predict`.

The new assertion here, that the parameter vector should always be 8 elements, is catching some problems.

Farnazmdi:
Thank you very much. I will look into the issues.

src/ODEmodel.jl
Farnazmdi: In this new setting, we don't estimate %G1 as a parameter anymore, right?

aarmey: Right. The control condition defines this for us.

Farnazmdi (approved):

PR #416, opened 2021-02-12
changes

Farnazmdi:


Farnazmdi (commented):
@aarmey I have been trying to run the fitting with the new structure. The cost function is reasonable and it finds a set of parameters, but when I plot it, it looks like this:
image
I printed out `v` and for every time point, the vector is zero (1e-18, 1e-16,...). Should we have `u = similar(v)`? I mean then we are multiplying `mul!(u, A, v)` and the `v` is zero. Does this line need modification?

Farnazmdi:
I tried to comment on line 79 in ODEmodel.jl, but it doesn't show. 
I meant in the `predict()` function.

aarmey:
That multiplication method is just to avoid allocation—it multiplies and puts the result in u, then the next line copies it over to v. The predict method is *exactly* the same between fitting and plotting. If you're getting reasonable cost values, I really think there has to be an issue in which parameters you are passing in.

PR #419, opened 2021-02-17
Find control through eigenvecs

aarmey:


src/ODEmodel.jl
```diff
@@ -28,25 +28,24 @@ function ODEjac(p::AbstractVector{T}, t::Real)::Matrix{T} where {T <: Real}
...
-    A = ODEjac(p, 100.0)
-    u = similar(v)
-
-    for ii = 1:100
-        v /= sum(v)
-        mul!(u, A, v)
-        copyto!(v, u)
-    end
+    @assert all(p .>= 0.0)
+    @assert all(p[7:end] .== 0.0) # No cell death in the control
```
aarmey, line 37: @Farnazmdi in the process of adding a more elegant solver for the control condition, I added this assertion that the control condition has no cell death. This is catching during the fitting, possibly indicating an indexing error somewhere. Could you take a look?

Farnazmdi: Of course, will do. Thank you.

Farnazmdi: It seems they are not exactly zero, but close.
image
I don't know where this happens, but we don't estimate the minimum of cell death, we just set them to zeros here:
image
Could we assert that they are *close* to zeros?

aarmey: I'm not sure what close would be. If they're evaluated at 0 concentration, and E_control is 0 for death, why aren't they 0?

Farnazmdi: Ahh.. That's why then. The concentration is not zero, because we want to get the log of it, we cannot have absolute zero. I had set the control to be a small number like 0.1. I can fix this right now in my branch so that we can have the concentration of control at zero in this function.

aarmey: Isn't the log of concentration only used for plotting? You should only do that correction before plotting, if so.

Farnazmdi: I will make sure that happens only in plotting.

Farnazmdi: This problem should be fixed.

PR #421, opened 2021-02-17
Test

Farnazmdi:
@aarmey does this work?

Please know that the only correctly-estimated set of parameters are for drug pairs fitting. For tests and notebooks that are based on joint fitting of all 5 drugs, or single drugs, I have not run the model and just replaced the parameters in the tests and notebooks with random vectors so they just run.


aarmey:
I'm ok with a hack like this if you put a TODO note and plan on eventually removing it. There's just too much risk otherwise of hidden errors in complex code like this. We should be able to satisfy any properties of the model we know to be consistently true.

PR #429, opened 2021-03-10
Multi plots for figure creation

Farnazmdi:
@aarmey  I was able to make figure 1. I will figure out how to append the cartoon of the model to it later.
But the build fails because of `makefile`. Could you please review it and let me know if I made a mistake in `makefile`?
Thank you.
image


makefile
```diff
@@ -6,5 +6,17 @@ all: $(patsubst %.jmd, %.pdf, $(wildcard *.jmd))
 	julia -e 'using Pkg; Pkg.activate("."); Pkg.instantiate(); ENV["GKSwstype"]="100"; using Weave; weave("$<", doctype = "md2pdf")'
 	rm $*.log $*.out $*.tex $*.aux
```
cyrillustan, line 8: Add `figure%.svg` to the first line, after `all`

cyrillustan:
Maybe check why this branch does not build?

aarmey:
@Farnazmdi I'll look at the build error. This figure needs to be completely restructured:

- All the information needs to be shown in a more compact form. For example, you've plotted different drug concentrations as varied shades on a single plot. That way you could just have two plots per drug (total cell #, and %G1).
- The experimental uncertainty should be shown. If the model fit is a line perhaps the experimental data can be distinguished by being a shaded region.
- The legends should be cleaned up, with consistent capitalization, limited abbreviations, etc. The caption shouldn't define things with duplication. For example, define how drug concentration is shown, and then define how to find experimental data vs. model fit. Don't make a legend item for every model/expt-concentration pair.
- The text should be larger. This is usually accomplished by making the dimensions of the figure smaller.

aarmey:
See my last commit for the fix. The `all` list states the files you want to build. It cannot include wildcards.

aarmey:
Last thing—these *have* to be consolidated because this figure also needs to summarize what the fit parameters are.

Farnazmdi:
@aarmey figure 1 now looks like this. The subplot (a) is the place for the model cartoon. Please let me know if you want any part of it changed. Thank you.
image

aarmey:
A few thoughts:

1. This needs to be horizontal rather than vertical to efficiently use space on a page.
2. Why is the range in (g) such that you can't see any of the changes?
3. It's confusing to have the pan-drug plots mixed in with the drug-specific ones.
4. Why is the starting G1% of all the model fits at 50%, if all of the data is higher?
5. Subplot labels should be the same font style as everything else (e.g., Helvetica)
6. The colors in the overlay are hard to distinguish and the contrast small... not sure how to make this better. An ordinal color series would help to interpret the plot.
7. Maybe G1# and G2# is clearer than total number and percentage?
8. Why is there an "f" in every plot?

Farnazmdi:
Thank you for the comments. I applied almost all of them.
Re 2. It is because none of the drugs have much of a G1 cell death effect. I wanted the G1 and G2 cell death plots to have the same y axis so I can't reduce that to less than 0.6. 
Re 3. I agree, should I add another figure for this? I can't think of a way to squeeze them into less number of subplots. Please let me know if you have a suggestion.
Re 4. It is strange for me too, but they are from the fitting.
Re 6. By overlay, you mean the time series plots? I removed one of the curves. Now it shows control and 2 other concentrations. You meant these colors don't seem distinguishable?
Re 8. It was the legend title, I removed it.
image

aarmey:
> Re 4. It is strange for me too, but they are from the fitting.

A systematic error like that is a strong signal to me that something might be wrong. Have you checked that you can get starting points other than 50% G1?

aarmey:
> Re 3. I agree, should I add another figure for this? I can't think of a way to squeeze them into less number of subplots. Please let me know if you have a suggestion.

Think about how you are going to write about these results. For example, the drug fits will come before you talk about the parameter values. The figures need to reflect this order.

If you don't think you'll have the space, show a few of the drugs in the main figure, and move the others to a supplementary one.

aarmey:
> Re 6. By overlay, you mean the time series plots? I removed one of the curves. Now it shows control and 2 other concentrations. You meant these colors don't seem distinguishable?

You can't just not show some of the data—add back all the concentrations. The colors need to be (1) distinguishable, (2) intuitively reflect the data. For example, Sean's figure where increasing drug concentration is shown by progressively darker lines is great. Right now you're using a categorial color series when something that reflects order will be more intuitive.

Farnazmdi:
update:
I have been struggling in controlling the color of `Geom.ribbon`. I can control the color of lines that show the model, but not the color of the standard deviation of data that we plot on top of it.
Besides that, I think having 8 lines and ribbons on the same plot will be really confusing.

Farnazmdi:
The reason we switched to Gadfly is that it provides higher resolution? 
Because we can plot multiple subplots in the same figure using `Plots` and it seems more stable and much better documented than Gadfly and I already had different types of plots working.

aarmey:
You can absolutely use plots. The only difference is in the interface.

PR #430, opened 2021-03-17
Figs

Farnazmdi:
Figure1 w/o the cartoon:
image

FigureS1 w/o the cartoon:
image

Please let me know if anything needs to be changed for now @aarmey.
Thank you.


Farnazmdi:
With few adjustments:
image

aarmey:
Shouldn't the other two drugs be in the supplement?

aarmey:
Also, check that your labels have consistent capitalization.

aarmey (changes requested):
You have empty files here. Are those needed?

Also, please don't commit the svg's.

Farnazmdi:
Should I put the other 3 drugs in the same figure as the simple_ODE? Or they could be separate? I mean space-wise, do we have enough space to have them as a separate figure?
The empty subfigures are for the cartoons of the model.
Of  course, I will remove the svg files,

aarmey:
You have space in the supplement. The simple ODE model should come after the full model for the other drugs.

Farnazmdi:
I made figureS1 as the rest of the model-data comparison:
image
The figure for simple ODE is now figure S2.
I fixed the capitalization and removed the svg files.
Could I merge this?

aarmey:
You can't just split these into more figures. Most journals require that there be at most one supplementary figure per main figure.

aarmey (changes requested):
What are the .jld files?

Farnazmdi:
They are the G1 and G2 cell numbers for the simple ODE model (the reference model) brcause I had to write a different ODEjac() and predict() so I decided to just save the cell numbers and load them for plotting the figureS2.

aarmey:
Ok, we can merge this after the figure changes, but I would strongly advise against doing this. If those numbers need to be changed, then you will have to recreate the model from scratch. It also hides the relationship between the results and underlying code.

Farnazmdi:
Oh I thought when you said the figure for simple model should come after the other one, it means as another figure.
I will put them together.

Farnazmdi:
I modified the figures from this last commit in Affinity designer, and this is the result. 
image

image

Please let me know if you want anything changed. Otherwise, I will send them to you in Slack.

aarmey:
Ok, this can be merged. I don't need the figure in Slack. You can start on the combination figure.

PR #432, opened 2021-03-18
Fig combo

Farnazmdi:
Figure 2 which should provide combinations, looks like this now:
image

I was trying to fix the subplot labels for hours yesterday, but couldn't figure out what's wrong with it! It plots the labels of the second half, on the plots of the first half!
I can fix it in Affinity designer if it is okay.
And please let me know if the structure is good or not.
Thanks.


src/figures/figure2.jl
```diff
@@ -0,0 +1,65 @@
...
+    plot!(t, g2_Bliss_model1, label=label2, titlefont = Plots.font("Helvetica", 12), lw=2, color="mediumpurple2", guidefont=Plots.font("Helvetica", 12), xtickfont=Plots.font("Helvetica", 12),ytickfont=Plots.font("Helvetica", 12), bottom_margin=1.5cm, fg_legend = :transparent, top_margin=1.25cm, left_margin=1.25cm, right_margin=1.25cm)
+    plot!(t, meanGS2[1, 1:189, i], label="G1 exp", titlefont = Plots.font("Helvetica", 12), lw=2, color="darkgreen", guidefont=Plots.font("Helvetica", 12), xtickfont=Plots.font("Helvetica", 12),ytickfont=Plots.font("Helvetica", 12), bottom_margin=1.5cm, fg_legend = :transparent, top_margin=1.25cm, left_margin=1.25cm, right_margin=1.25cm)
+    plot!(t, meanGS2[2, 1:189, i], label="G2 exp", titlefont = Plots.font("Helvetica", 12), lw=2, color="mediumpurple4", guidefont=Plots.font("Helvetica", 12), xtickfont=Plots.font("Helvetica", 12),ytickfont=Plots.font("Helvetica", 12), bottom_margin=1.5cm, fg_legend = :transparent, top_margin=1.25cm, left_margin=1.25cm, right_margin=1.25cm)
+    annotate!(-25.0, 87.0, text(subPlabel, :black, :left, Plots.font("Helvetica Bold", 15)))
```
aarmey, line 9: The y-position here is way too high. That's why they're on the wrong plot.

aarmey:
We need to compare our model with a baseline, such as Bliss on the cell numbers. Where are we going to show that?

Farnazmdi:
The first row shows the comparison between the experimental data and the "Bliss on cell numbers" that I called it "Bliss on data".
The second row shows the comparison between the experimental data and "Bliss on model".
I know the legends are a bit confusing. I couldn't come up with a better-short-word for calling those two.

src/figures/figure2.jl
Farnazmdi: Thanks! Fixed.

Farnazmdi: image

Farnazmdi:
I had a hard time putting together the "Bliss on cell number", "Bliss on model", and "experimental data". It took me longer to put them together because of complicated indexing...
Here is figure 2:
image

And figure S2 where I plotted a few other plots in the same structure as figure2. Since there are a lot of them, I thought the SSE plot may suffice. 
image

Please let me know if you want me to keep or discard figure S2. And that if figure2 is good. I will fix the few details, such as slightly-not-aligned titles/labels in affinity designer.
Thank you.

aarmey:
What's going on with the error bars in (d) and (h)?

This seems close to something we can send to Sean and Laura for feedback.

Farnazmdi:
I fixed it, the problem was that in that exact condition, (gem 30 & lapatinib 100) we only had 2 replicates, not 3 and it was calculating the average of those two replicates with zero as the third replicate.
image

PR #449, opened 2021-04-15
Erlang

Farnazmdi:
To plot G1 and S/G2 distributions of control conditions (fit Erlang).
A part of Figure S2.

@aarmey  I had fitted the data in Python, but now that I tried in Julia the numbers are not the same...
Here is the estimated shape and scale for control condition in G1 for all 5 drugs, in order: (they are pretty consistent!)

image

And for S/G2:

image

This says the number of G1 subphases is ~12 and S/G2 is ~ 5
which is a bit strange as we expect the G2 subphases to be greater.

Should I change the structure of the model according to these?

Another observation:
I looked into different concentrations of lapatinib: For data in higher concentrations, in G1, the shape gets larger and goes up to ~ 70. For G2, the scale parameter changes, and the shape is between 5 and 10.


aarmey:
There is no reason there should be a Python/Julia difference. Are there other explanations for the difference?

Farnazmdi:
Yes, in Julia I just used a built-in package, but in python I had written a function that manually was calculating the estimated parameters, like the closed form estimator we have in tHMM. 
In that, we loop over integer values for shape and use the average of data inside the function and pick the one with the highest likelihood.

aarmey:
What is it you are trying to do here?

Farnazmdi:
The Erlang distribution fits (for control conditions) that will go to supplementary figure 2.

aarmey:
And what is the point of that plot?

Farnazmdi:
To show the data fits Erlang and that we picked the number of subphases according to the estimated shape parameter from the data.

Farnazmdi:
Sorry, my bad-- I made a mistake and was working on the wrong data...

aarmey:
We didn't chose the number of subphases to exactly match the fit—we made this change after M&M meeting.

Indeed, it's to show the times follow an Erlang distribution. So you can fit using the existing estimator then overlay the fit on the data.

Farnazmdi:
Will do.
Thank you.

Farnazmdi:
It was actually pretty close... I was fitting the wrong data..
The figure now looks like this:
image

aarmey:
You need to show the distribution as a line, not a histogram of samples.

Farnazmdi:
I spent more than half a day finding how to plot the line instead of a histogram for the distribution but it always normalizes the density... I will search more.

aarmey:
Can't you make a vector of f(x) from the distribution, then multiply all the values by a number?

Farnazmdi:
I will try it. Thank you

Farnazmdi:
Worked perfectly! Thank you
image

PR #452, opened 2021-04-24
combo fitting start

Farnazmdi:
I organized the data for the drug combinations and started writing functions for fitting.
One problem I have is the concentrations. In the Hill function, we need to have the value of concentrations, but what would be the value for the combination of two drugs? Say 100 nM lapatinib + 30 nM Gemcitabine? It must be important because the model estimates the EC50.
Do you have any suggestions for this @aarmey? Should I just sum them up?


aarmey:
You use the Hill curve for each of the drugs separately, then Bliss tells you how to combine their effects. You already do exactly this when you make combination predictions from the model, do you not?

aarmey (commented):
What is the .jld binary for?

Farnazmdi:
I thought I am supposed to fit the combination time series and find the combined effects straight. 
I re-organized the data file so the fitting would be easier, such that we have [control, single_drugA, drugA+drugB1, drugA+drugB2, drugA+drugB3, drugA+drugB4] as the conditions. And I saved them as JLD files so accessing would be easier.

I actually ran it (just summed up the concentrations in combinations) but the fits are not great.

I guess my approach is wrong. I will try and figure out your suggestion.

aarmey:
What do you mean by "fit the combination time series and find the combined effects straight"? That sounds like doing what I suggested, but then you said you couldn't do that?

src/figures/figureS5.jl
```diff
@@ -0,0 +1,96 @@
...
+########### 5. control, Doxorubicin 20, Doxorubicin 20 nM + gemcitabines [5 nM, 10 nM, 17 nM, 30 nM]
+########### 6. control, Lapatinib  100, Lapatinib  100 nM + palbociclibs [25 nM, 100 nM, 250 nM, 250 nM]
+
+concens = zeros(6, 6)
+concens[:, 1] = [0, 50, 75, 100, 150, 300]
+concens[:, 2] = [0, 50, 55, 60, 67, 80]
+concens[:, 3] = [0, 10, 35, 110, 260, 260]
+concens[:, 4] = [0, 10, 35, 60, 110, 260]
+concens[:, 5] = [0, 20, 25, 30, 37, 50]
+concens[:, 6] = [0, 100, 125, 200, 350, 350]
```
Farnazmdi, line 16: I meant that we treat the combination time series the way we did the single drug treatments. For example in case of palbo 50 + gem 10, I didn't know what I should set as the concentration, so I summed 50 + 10. 
I extracted 6 types of combinations that are listed here on the top, and the `concens` shows the concentrations I used for these conditions. But I guess you meant we should extract single drug treatments from fitting the combination data? I imagine I should dd another step to the fitting and pass single drug parameters through Bliss and the combined effect will be the one used for fitting?

aarmey:
Yeah, you can't just add the two concentrations together—they're completely different compounds.

Since drugA is at a constant concentration, you could just use the Hill effect with respect to drugB.

Farnazmdi:
I only tried palbo 50 + lapatinibs [25 nM, 50 nM, 100 nM, 250 nM]. I wanted to make sure this is the correct way, then I will fit all together. Although, for some drugs -- like dox -- we only have 1 concentration, and for some others we have 2, or 3, or 4 concentrations. I guess it makes it hard to have an integrated set of parameters, however, I can do exactly what I did here for other 5 conditions I have. Do you mind reviewing this PR and letting me know if this is the correct way?

image

Farnazmdi:
@aarmey I fitted the combination with and without the ones including DOX (we only had dox 20 nM combined with other drugs and no other concentration) 

image

image

There is only one control condition, so we don't have that problem, but still, some of the fits, like gem with palbo, are not good enough.
Should I proceed with it or keep the single-drug fits as before?

aarmey:
I don't understand what you are doing in the most recent results. Just like the previous, but all the combinations share the same control?

aarmey:
In your results from a few days ago, how do the effects of Lapatinib in the presence of palbo compare to those with only Lapatinib?

aarmey:
Also from 3 days ago—I'm not sure how you are able to fit both the control and palbo only? Since you only have the Hill curve relationship for Lapatinib, wouldn't palbo alone be the control, and then the no-drug condition wouldn't be included?

aarmey:
I guess mainly I could use a clear explanation of exactly what you have done here.

Farnazmdi:
Absolutely, sorry, should have explained it before asking you to review it.
In the combinations, we have these sets of data:
`x1 = ["control" "palbo 50 nM" "palbo 50 nM + lap25 nM" "palbo 50 nM + lap 50 nM" "palbo 50 nM + lap 100 nM" "palbo 50 nM + lap 250 nM"]`
`x2 = ["control" "palbo 50 nM" "palbo 50 nM + gem 5 nM" "palbo 50 nM + gem 10 nM" "palbo 50 nM + gem 17 nM" "palbo 50 nM + gem 30 nM"]`
`x3 = ["control" "gem 10 nM" "gem 10 nM + palbo 25 nM" "gem 10 nM + palbo 50 nM" "gem 10 nM + palbo 100 nM" "gem 10 nM + palbo 250 nM"]`
`x4 = ["control" "gem 10 nM" "gem 10 nM + lap 25 nM" "gem 10 nM + lap 50 nM" "gem 10 nM + lap 100 nM" "gem 10 nM + lap 250 nM"]`
`x5 = ["control" "dox 20 nM" "dox 20 nM + gem 5 nM" "dox 20 nM + gem 10 nM" "dox 20 nM + gem 17 nM" "dox 20 nM + gem 30 nM"]`
`x6 = ["control" "lap 100 nM" "lap 100 nM + palbo 25 nM" "lap 100 nM + palbo 50 nM" "lap 100 nM + palbo 100 nM" "lap 100 nM + palbo 250 nM"]`

The control is the same for all, and we have those single-drug treatments as well.
let's ignore `x5` because we only have a combination of dox20 with others and no other concentration of dox. In contrast, we have 4 concentrations of lapatinib, gemcitabine, and palbociclib (combined with each other in pairs). For these 4 concentrations, we can define a hill for either of these three drugs and that's what I did.

Basically, we estimate the hill parameters for single drugs for 4 concentrations, and we pass them through Bliss to find the Bliss-combined parameters -- for those pairs that we have data -- and then use these parameters to fit the data to the model, using `predict()` that outputs the cost function.
We estimate 48 parameters (which are the single drug parameters for 3 drugs) through which we fitted the bliss-combination of them to the combination data.

* I only used the data from the new experiment.

Re: "In your results from a few days ago, how do the effects of Lapatinib in the presence of palbo compare to those with only Lapatinib?"I should have added the lapatinib-only, too. I will as soon as you approve of my approach in here. The results from a few days ago, was only `x1` where I didn't assume a hill for palbo, and just assumed we have ODE parameters to estimate.

Re: "Also from 3 days ago—I'm not sure how you are able to fit both the control and palbo only? Since you only have the Hill curve relationship for Lapatinib, wouldn't palbo alone be the control, and then the no-drug condition wouldn't be included?" 
I have both in here. Since in the new experiment, we had single drug treatments for *some* of the conditions. For example, since we have combination of palbo 50 with lapatinibs, we had palbo-only and I used it.

Please let me know if I didn't explain clearly.
I will also comment on the functions to show what I have added.

src/figures/figureS4.jl
```diff
@@ -201,14 +217,19 @@ lap100gem[1, :, 2], lap100gem[2, :, 2], _ = predict(LPT_GEM[:, 6, 7], LPT_GEM[:,
...
 
-########### Lap 100 nM + palbociclibs [25 nM, 100 nM, 250 nM]
+########### Lap 100 nM + palbociclibs [25 nM, 50 nM, 100 nM, 250 nM]
 Lap100palbo = zeros(3, 189, 3)
 for i = 2:3
     Lap100palbo[1, :, i], Lap100palbo[2, :, i], _ = predict(LPT_PLB[:, 6, i + 4], LPT_PLB[:, 1, 1], t)
 end
 Lap100palbo[1, :, 1], Lap100palbo[2, :, 1], _ = predict(LPT_PLB[:, 6, 4], LPT_PLB[:, 1, 1], t)
 Lap100palbo[3, :, :] .= Lap100palbo[1, :, :] .+ Lap100palbo[2, :, :]
-# well 2: 8, 14, 20
```
Farnazmdi, line 211: All the changes in `figureS4.jl` are to re-organize the data. Please ignore them.

src/fitCombo.jl
```diff
@@ -0,0 +1,72 @@
...
+function RESID_through_bliss(p, g1c, g2c)
+    p_combo = my_helper(p)
+    t = LinRange(0.0, 0.5 * size(g1c, 1), size(g1c, 1))
+    res = 0
+    for j=1:5
+        for i=1:6
+            res += predict(p_combo[:, i, j], p_combo[:, 1, j], t, g1c[:, i, j], g2c[:, i, j])[1]
+        end
+    end
+    res
```
Farnazmdi, line 20: Here is the residual function. We have 5 pairs of combinations in the data, and 6 concentrations [control, drugA, drugA+DrugB1, drugA+DrugB2, drugA+DrugB3, drugA+DrugB4]

src/fitCombo.jl
```diff
@@ -0,0 +1,72 @@
...
+end
+
+""" Unit function to calculate the bliss for 2 drugs at one specific concentration. """
+function Bliss_unit(pp1, pp2, control)
+    # pp1 and pp2 are 1D arrays of size 8, including 8 parameters for a single concentration.
+    p1 = copy(pp1)
+    p2 = copy(pp2)
+    # normalization
+    p1[1:6] .= 1.0 .- (pp1[1:6] ./ control[1:6]) # g1 and g2 prog. rates
+    p1[7:12] .= pp1[7:12]                          # g1 and g2 death rates
```
Farnazmdi, line 41: I only added this function, because we have 1 control condition. In the original code, since each single drug had its own control, I used to get the average of them when normalizing and scaling back after calculating combination, which is not necessary here.

src/fitCombo.jl
```diff
@@ -0,0 +1,72 @@
...
+    dox = p[1:12]
+    p_ode = getODEparams(p, concs) # returns 12 x 5 x 3
+    p_combo = zeros(12, 6, 5)
+    p_combo[:, 1, :] .= p_ode[:, 1, 1] # control
+
+    for i=2:6
+        p_combo[:, i, 1] = Bliss_unit(p_ode[:, 3, 3], p_ode[:, i-1, 1], p_ode[:, 1, 1])
+        p_combo[:, i, 2] = Bliss_unit(p_ode[:, 3, 3], p_ode[:, i-1, 2], p_ode[:, 1, 1])
+        p_combo[:, i, 3] = Bliss_unit(p_ode[:, 2, 2], p_ode[:, i-1, 3], p_ode[:, 1, 1])
+        p_combo[:, i, 4] = Bliss_unit(p_ode[:, 2, 2], p_ode[:, i-1, 1], p_ode[:, 1, 1])
```
Farnazmdi, line 67: This is basically a `getODEparams()` for calculating the combined parameters. Converting hill to ODE params.

aarmey:
Ok, thanks. This does sound like a reasonable approach. It looks like some fit very well and then others fit very poorly. Would you agree? It might help to investigate the ones that fit most poorly, and see if that is because the response is very different than what we would expect (by just looking at the plots).

Farnazmdi:
Yes, I agree completely. Those fits that have gemcitabine involved are poor. I will plot the parameters of control, single-drug treatments, and Emax_combination and post it here. Also, will look into the time series plots.

Farnazmdi:
I plotted the single drug treatment data from the previous experiment along with simulations from fitting combination data. This only shows for gemcitabine + palbociclib that the fits were not good. 
It seems to me that the model can follow the *pattern* but does not fit exactly and needs some sort of re-scaling. 

image

aarmey:
Doesn't seem like this is fitting the combination data all that well, either. Was this a "good" combination fit case?

Farnazmdi:
It was the "bad" combination case.
This is a good fit:
image

Farnazmdi:
I will merge this branch (it has been open for a month) and work on other changes in a new branch.

PR #458, opened 2021-05-14
using JuMP

Farnazmdi:


src/Hill.jl
```diff
@@ -37,25 +36,33 @@ end
...
-        nMax = 1
-    elseif length(p) == 34
-        nMax = 2
-    end
+function getODEparams(p, concs)
+
+    nMax = Int((length(p) - 6) / 14)
 
-    effects = zeros(eltype(p), 12, length(conc[:, 1]), Int((length(p) - 6) / 14))
+    effects = zeros(eltype(p), 12, length(conc[:, 1]), nMax)
```
Farnazmdi: I have set it up here, but it has `type` problems that I can't make sense of it. 
This is the error:

image

I also tried fitting for only the control condition, but it threw another `type` error for `startV()`.

aarmey: Why do you need JuMP? I don’t know anything about how to use it.

aarmey: If it’s because Optim.jl doesn’t have the right constraints, I’d start by just adding a penalty in the cost function for the region you want to avoid.

Farnazmdi: Yes, that's why I set this up. Thank you, sure, I will look into that and implement it in this branch.

PR #459, opened 2021-05-15
fit

Farnazmdi:
Thank you for the suggestion, it is a nice trick.
The cost function is now modified this way:
`new_cost = old_cost + 60*(maximum[0, (x_max - x_min)])^2`
I had to increase the max_iter to 800000 to get a good fit.
Here is the new simulation and parameters:

image

image


Farnazmdi:
This is fitting single_drug_treatments_atonce. 
I added a `10 * (max_effect - control) ^ 2` to the cost function, but it doesn't seem to work. The fits are good, but still progression rates can increase! Am I doing it wrong or factor 10 is small? Is there any modification that you think can help? @aarmey 

Here is the simulations 
image


image

Farnazmdi:
Update:
I increased 10 to 100 and it is better. I will make it 500 and see if it gets even better or not.

aarmey:
What is your conclusion about the fitting quality?

Farnazmdi:
I think it is worth having the constraints. The fits are not drastically worse, the difference is very little.

aarmey:
Agreed. Merge and I can take a look at the revised figures when you have them.

Farnazmdi:
Will do, thanks. I am also running the combination fits with this modification. I will update you when I have the results.

PR #481, opened 2021-07-04
new parameters

Farnazmdi:
I made the changes to the parameters without changing the jacobian matrix, by defining a new function as a sort of filter that the new parameters that we estimate, pass through that and will look like the old parameters and everything would look the same.

The bad news is the fits are not better:
image

image

I tried widening the range of parameters 5 times, but it did not improve the fitting.

One other way to make another change to the model is changing the number of steps. Instead of G1 having 2, and G2 having 4, I can try 3 and 3 or 2 and 2 and see if it gets better. 
This way it is a little easier to explain why we chose these, for example, to distinguish the effects into the "first and second half" (2 parts), or into "start", "middle" and "end" of the phase (3 parts).

Please let me know if you think I should try these or we just keep the old setup.


aarmey:
Interesting... how did you handle the death probability with respect to concentration? Is it a Hill curve with range 0–1?

I agree that splitting the phases to 3 G1/3 G2 would be a reasonable next step. Our justification for a different number in G1/G2 has always felt awkward to me, anyway.

Farnazmdi:
@aarmey Thanks, I will do that next.
The death probability does not have a specific form, at least I did not set it to have. I defined a new "death rate" by multiplying the `"rate" * (1 - p)` where `p` is the probability of transition to the next step. In the same way, the "transition rate" is defined by `"rate" * p`. 
So right now, the input parameters that are being optimized, are the "rate"s and "p" and I defined a function to convert these into the previous form of transition rates and death rates. This way, we don't need to change the jacobian matrix.
Does this make sense to you?

aarmey:
So you are converting away and then right back to the original parameterization? It's no surprise that it would fit exactly as well in that case—it's the same model! The point is to try and fit using the "probability" parameterization, as this provides a different view of drug dose response relative to the "rate" parameterization.

Farnazmdi:
image

image

It does not seem improved...

aarmey:
Ergh. Ok, let's try splitting up the cell cycle differently then...

Farnazmdi:
Since this approach did not improve fitting I am closing this PR.

PR #482, opened 2021-07-07
Three

Farnazmdi:
@aarmey Could you help me please? Since last night I have had this problem. It seems like there is more than 1 positive eigenvalue for the control condition when it starts running the fitting. Should I choose the one that is larger or you think there is an error?

image


aarmey:
Have you looked at what they are when they are problematic?

Farnazmdi:
These are the three components that are positive. 
0.0003772975143800164 - 0.2389237832482493im, 
0.0003772975143800164 + 0.2389237832482493im, 
0.026237394679327564 + 0.0im

aarmey:
These should work fine, because only one is positive AND real.

Farnazmdi:
Did not improve it that much :'(
Are all the projects like this? Nothing makes the fits better.. It is surprisingly similar to the one we already have with 2 and 4 steps.
image

image

aarmey:
How difficult is it to try four and four? We did make G2 simpler...

This has been a devilishly difficult project. :-) Hopefully we can be done with it soon.

Farnazmdi:
It is not difficult to do the 4 by 4. I will do it today. 
Thank you!  I hope so : ) 
I hope after all these the fits get better...

Farnazmdi:
One thing I noticed, is that for gemcitabine, the restriction we forced about the progression rates to be strictly decreasing, is affecting the fits. Could we relax that restriction only for gemcitabine? The G1 durations -- based on experimental data -- are not extended with higher concentrations: so I think it is reasonable to do that.
image

aarmey:
Sure.

Farnazmdi:
For 4 and 4 steps, when estimating the control condition, we don't have  eigenvalues; only complex conjugates and the last pair have positive real components:
0.02011302861560011 - 0.13149662289402428im, 
0.02011302861560011 + 0.13149662289402428im

I understand we are assuming pure exponential for the control and we should have a pure real eigenvalue for it. I don't know what to do in this situation. Should I keep the last one?

aarmey:
Hmm... whenever I had trouble with the eigenvalues before, it turned out to be a parameterization issue. Presumably this is a rare issue? If so, are the parameters that cause this different in any way from the ones that work?

Farnazmdi:
The parameters are not different, only there are more parameters.. You are right it might be something I missed or a mistake when I was writing the Jacobian. I will check again. Thanks.

aarmey:
Great. Indeed, take a second look, and let me know if you can't find anything. Does it happen with every solve, or occasionally?

Farnazmdi:
Will do. Thank you. I thas happened every time I ran it.

Farnazmdi:
You were right! I had an extra `-` in front of a set of parameters.. It is working now. I will update you about the fitting here. Thanks!

Farnazmdi:
The fits are better now. I will merge this PR to merge the changes to my other branches.

PR #507, opened 2021-11-30
importing new data

Farnazmdi:
Probably not the most efficient way of importing, but I managed to put the conditions related close to each other (similar to the way we had the previous data).
There are 3 or 4 replicates from which I have not got the average yet.
I will try the fitting with this.


src/importData.jl
```diff
@@ -54,6 +54,141 @@ function import_combination(filename::String)
...
+    basePath = joinpath(dirname(pathof(DrugResponseModel)), "..", "data")
+
+    d = zeros(193, 1248, 2) # time points x conditions x G1% and total
+
+    dataname = "AU565_round2and3_model_subset_level2.csv"
+    data = readdlm(joinpath(basePath, dataname), ',')
+
+    conditions = Vector{String}()
+
+    for i = 1:96 # first set that have 191 data points
```
aarmey, line 73: All this looping should be a table [reshape](https://dataframes.juliadata.org/stable/man/reshaping_and_pivoting/).

src/importData.jl
```diff
@@ -54,6 +54,141 @@ function import_combination(filename::String)
...
+    end
+    for i=96:1247 # the rest that have 193 data points
+        d[:, i+1, 1] = data[(193 *(i-1)) + 3 : (193 * i) + 2, 9]
+        d[:, i+1, 2] = data[(193 *(i-1)) + 3 : (193 * i) + 2, 8]
+        push!(conditions, data[(193 *(i-1)) + 3, 5])
+    end
+
+    # filter
+    for i = 1:size(d, 2)
+        d[:, i, 1] = savitzky_golay_filter(d[:, i, 1], 41, 3)
```
aarmey, line 86: I'd suggest having a function that takes in a vector and does this filtering so that the filter settings are defined in one place.

src/importData.jl
```diff
@@ -54,6 +54,141 @@ function import_combination(filename::String)
...
+    # separate G1 and SG2 cell#
+    gs = zeros(2, size(d, 1), size(d, 2)) # G1# , G2# x time x condition
+    population = init_cells * d[:, :, 2]
+    gs[1, :, :] = population .* d[:, :, 1] # G1
+    gs[2, :, :] = population - gs[1, :, :] # G2
+
+    return gs, conditions
+end
+
+function reorganize(gs, conditions)
```
aarmey, line 98: What is going on here?!

src/importData.jl
```diff
@@ -54,6 +54,141 @@ function import_combination(filename::String)
...
+    end
+    control = reshape(control, 193, c); FU = reshape(FU, 193, f); pano = reshape(pano, 193, p); AZD = reshape(AZD, 193, a)
+    BEZ = reshape(BEZ, 193, bz); Bort = reshape(Bort, 193, b); MG = reshape(MG, 193, m); Ever = reshape(Ever, 193, e)
+    MK = reshape(MK, 193, mk); JQ1 = reshape(JQ1, 193, j); Tram = reshape(Tram, 193, t); Cabo = reshape(Cabo, 193, cb)
+    new_gs = hcat(control, FU, pano, AZD, BEZ, Bort, MG, Ever, MK, JQ1, Tram, Cabo)
+    @assert(size(gs) == size(new_gs))
+    return new_gs, new_c
+end
+
+function reorganize_data()
```
aarmey, line 181: What is the purpose of this function?

aarmey (changes requested):

Farnazmdi:
Here is just the data:
image
image

Here is the total cell numbers:
image
image

I used Cyrillus's `figureCommon.jl` in the FCRegression repo, I hope it is alright with you. The figures look much better!

aarmey:
Great! Make sure the y-axes match?

Farnazmdi:
The axes are now from 0.0 to 2.0. 
I also edited that importing function. Now, given the name of the drug, we can call the g1, sg2, and the concentrations for all replicates of that drug.
I used 4 replicates of the vehicle as for the concentration 0 for all drugs, as Dr. Dane said the replicates are not exactly specific for each drug.
I believe this PR is ready to merge.

aarmey (commented):
`data/AU565_round2and3_model_subset_level2.csv`... is this being used here?

aarmey:
One question, otherwise this looks good.

Farnazmdi:
Yes, that is the new data. It is used in the `import_data()` function which is called in other functions.

src/figures/figure1.jl
```diff
@@ -0,0 +1,74 @@
...
+end
+
+function figure1()
+    setGadflyTheme()
+    g, c = DrugResponseModel.import_data()
+    newg = DrugResponseModel.trim_data(g, c)
+    tensor, conditions = DrugResponseModel.form_tensor(newg, c)
+    tens = mean(tensor, dims=2)
+
+    p1 = Eachdrug(tens[1, 1, :, :, 1], conditions[1], "G1")
```
aarmey, line 17: All of these should be handled as a loop.

src/figures/figure1.jl
```diff
@@ -0,0 +1,74 @@
...
+    return draw(SVG("figure1.svg", 18inch, 20inch), pl)
+end
+
+function figure2()
+    setGadflyTheme()
+    g, c = DrugResponseModel.import_data()
+    newg = DrugResponseModel.trim_data(g, c)
+    tensor, conditions = DrugResponseModel.form_tensor(newg, c)
+    tens = mean(tensor, dims=2)
+    p1 = Eachdrug(tens[1, 1, :, :, 1] .+ tens[2, 1, :, :, 1], conditions[1], "total")
```
aarmey, line 60: Should be handled as a loop.

aarmey (changes requested):
Ah, ok, I see that now.

In that case, what are the xlsx files for? I also found two places where the code should be simplified with loops.

Farnazmdi:
Oh.. my bad.. I thought I had deleted the xlsx files. I will do it right away.

aarmey (approved):
Excellent!

PR #508, opened 2021-12-13
figure2 showing model and fits

Farnazmdi:
@aarmey I used the average of 4 replicates to fit the model, and here are the results.


Farnazmdi:
image
image
Something strange I think is that the control condition seems to be oscillating..
I will check with the figure that Dr. Dane sent us to make sure the controls I plotted are actually control

Farnazmdi:
Oh.. yes, that's right! I used the vehicle, and not the control.
Should I change any assumptions that we had for control, or should I use control instead of the vehicle?

aarmey:
Well, I think first determine whether the control oscillates.

PR #509, opened 2021-12-16
Params

Farnazmdi:
Here are the plots I made. If any of them don't seem good or necessary, please let me know and I will remove them.
I know that for those PCA plots, it is better to do a CMTF instead of PCA since we have 8 concentrations, but I was not sure if we want to pursue this analysis.

image

image

image

image

image

image

image

image


aarmey (approved):
Great. Let's go through your individual fits first, to make sure they all look reasonable. There's not much significance to the PCA of the parameters if the parameters don't carry meaning.

Farnazmdi:
Of course, I will try to find out the phase effects of these drugs in the literature to compare, and if I could find something I will update you today during the meeting.

